PR #9, opened 2015-03-13
Fixing targets files to remove duplicate import warnings

smaillet-ms:
Another partial fix for issue #5. This fix updates the targets files to prevent duplicate MSBUILD file imports.


mortezag:
:+1:

PR #14, opened 2015-03-13
Eliminate or Suppress Linker warnings for Build_SDK

smaillet-ms:
Partial fix for issue #5 
* Eliminate Linker warning 4022
* Supress Linker warnings 4221, and 4199

- Added IGNORE properly to linker tasks and LinkIgnore project properties to set the warnings to ignore
4022 was due to duplicte C/CPP files in the list of files to compile or inclusino of a header file as a compiled source. These were fixed, the others are real but benign warnings and are no set to suppress.


mortezag:
:+1:

Issue #15, opened 2015-03-13
build_sdk fails if build_share is specified for a location that isn't a child of SPOCLIENT


smaillet-ms:
When running build_SDK in the build server the output of the build is redirected to a common build drop share. To do that, the build_share parameter is used to specify the location of the share. However, since several of the projects are copied (and trasnformed) and then built from the output folder the BuildEnv.props file won't be found, which generates an error.

smaillet-ms:
Fixed in pull #16

PR #16, opened 2015-03-13
make all imports of BuildEnv.props conditional 

smaillet-ms:
Fix for issue #15 
Made all imports of BuildEnv.props conditional as some of the project files are copied (and transformed) into the output location before being built, thus BuildEnv.props won't be in the path anywhere. This only occurs when doing an SDK build when the environment is already set up. BuildEnv.props isn't needed in that case, so not importing it is OK.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 # Build results
-BuildOutput/
+/BuildOutput
 bin/
 obj/
 *.cache
@@ -14,3 +14,4 @@ obj/
 
 # WinPCAP
 /DeviceCode/Targets/OS/Win32/DeviceCode/WinPcap_Eth/Dependencies/WpdPack/
+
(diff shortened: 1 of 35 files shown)


.gitignore
```diff
@@ -1,5 +1,5 @@
 # Build results
-BuildOutput/
+/BuildOutput
```
mortezag: What is the reason of this change?

mortezag:
:+1:

.gitignore
smaillet-ms: VS integration was showing the BuildOutput as a non-ignored folder. So I used VS to set the ignore and this is the syntax it uses. (It's a bit more correct as the intent is to ignore the %SPOCLIENT%\BuildOutput folder as opposed to any folder under %SPOCLIENT named BuildOutput)

PR #17, opened 2015-03-16
Fixed MMP error MMP0000: 0x80131700. Fixes #12

### LINKED ISSUE #12: MMP : error MMP0000: 0x80131700
MetaDataProcessor.exe fails with the above error during the SDK build on a computer where only .NET Framework 4.0 is installed (.NET 2.0/3.5 is not installed). It also happens during the build of a .NET Micro Framework project from within Visual Studio.

[NETMF CodePlex Issue [#221](https://netmf.codeplex.com/workitem/221)]

cw2:
This is a fix for issue #12.

Used .NET Framework 4.0 CLR Hosting Interfaces; NETMF does not require .NET 2.0/3.5 anymore.

### DIFF
--- a/CLR/Tools/MetaDataProcessor/MetaDataProcessor.vcproj
+++ b/CLR/Tools/MetaDataProcessor/MetaDataProcessor.vcproj
@@ -67,7 +67,7 @@
 			<Tool
 				Name="VCLinkerTool"
 				AdditionalOptions="/DEBUGTYPE:cv,fixup /nxcompat"
-				AdditionalDependencies="dbghelp.lib crypto.lib "
+				AdditionalDependencies="dbghelp.lib crypto.lib mscoree.lib"
 				OutputFile="$(OutDir)/MetaDataProcessor.exe"
 				LinkIncremental="0"
 				AdditionalLibraryDirectories="$(SolutionDir)Crypto\lib\x86"
@@ -157,7 +157,7 @@
 			<Tool
 				Name="VCLinkerTool"
 				AdditionalOptions="/DEBUGTYPE:cv,fixup /nxcompat"
-				AdditionalDependencies="crypto.lib"
+				AdditionalDependencies="crypto.lib mscoree.lib"
 				OutputFile="$(OutDir)/MetaDataProcessor.exe"
 				LinkIncremental="1"
 				AdditionalLibraryDirectories="$(SolutionDir)Crypto\lib\x86"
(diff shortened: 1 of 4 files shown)


miloush:
Is it necessary to have the runtime version hardcoded in the application? Looks a bit to me like postponing the problem to the next version...

smaillet-ms:
@miloush - I agree, can we do this with a default framework for compatibility but with the option to provide the version on the command line.

@cw2 - Please don't use #pragma comment to reference lib files. Use the command line linker lib inputs. Hiding dependencies with pragmas makes static dependency analysis much harder.

cw2:
@miloush Well, it is not strictly necessary; but the runtime version has to be specified at some point. Please consider this as a temporary fix for .NET 2.0/3.5 dependency only, until the MMP is rewritten in managed code ([has been already done](https://github.com/xamarin/Monkey.Robotics/tree/master/Source/Xamarin%20Studio%20Microframework%20Add-in/MFMetaDataProcessor) by David Karlas/Xamarin).

miloush:
Couldn't you find out what version is present and use that one (or use any available)? Interesting, I am actually working on that too. :-)

cw2:
@smaillet-ms Do you mean [adding mscoree.lib to] ExtraLibs in "tools\Targets\Microsoft.SPOT.System.x86.Targets"?

smaillet-ms:
No, I mean adding it to the list of libs for the metadataprocessor's project only - putting it in the x86 targets would do so for ALL x86 based projects in the build and this lib/change is only for ONE specific project

smaillet-ms:
@miloush - Definitely don't want to pick whatever is available, as that could have breaking implications. We want the build to be repeatable by anyone getting the code (and in particular by our build infastructure so we can go back to any version and get repeatable builds in the face of an audit of some sort). Picking whatever was present would mean you are potentially working from an untested, and possibly non-working version.

A given version of the framework should be prescriptive of the versions of all it's dependencies.

miloush:
@smaillet-ms It is probably always the choice of either consistently non-working solution with frustrated users or potential hard to reproduce failures. While I like the idea of command line override, is there a way to enforce it from Visual Studio? Anyway, I believe that as simple solution as providing better error message would help a lot, even if the version was left in hardcoded. New CLR versions are not being released that often to be fair.

smaillet-ms:
@miloush, it wouldn't be forced in VS, it would be set in the msbuild project files for a given version of the SDK, which could include dependency checks in the MSI. That way it is always locked to a well known working version and that dependency is made explicit at installation of the SDK. The issue this fix is for though is for the framework and firmware builds from GitHub sources. Which don't have an installer. Thus we need to have some explicit checking or, simply rely on the fact that we require a specific msbuild and VS installation for the build and that will include the correct version. The last thing we want to do is leave it to any arbitrary install that hasn't been tested and causes a break in less than obvious and subtle ways at runtime of the device.

PR #19, opened 2015-03-16
Fixing build break

smaillet-ms:
Fixing build breaks as previous commit removed PK.proj, when in fact it is needed to re-build the tools used that build everything else, and not the old porting kit itself. Eventually we will want to split this out so the tools can be built only when really needed and otherwise just used for a normal build.


mortezag:
:+1:

PR #20, opened 2015-03-16
Provide user feddback instead of indecipherable errors when CMSIS support files are missing

smaillet-ms:
Added InitialTargets to project files to check for CMSIS-Core and CMSIS-RTX pre-requsites for the CMSIS-RTX project (and therefore any solution that requires it) and the STM32F4.settings (and therefore any solutions that require it)


mortezag:
:+1:

PR #1, opened 2015-03-10
MCBSTM32F400 External FLASH and SSL support

smaillet-ms:
- Update SSL and sockets to support building with ARM MDK 5.0x tool chains (some macros used in OpenSSL conflict with MDK runtime names)
- Added support for MCBSTM32F400 External NOR chip to store Deplyment assemblies.
- Added Managed code Hardware project so the MCBSTM32F400 is closer to a real design (Goal is to deliver a golden reference of best practices with this platform, this is one of many steps to get to that goal)
- added missing CorDebug.sln for building/debugging VS integration suppport for VS2012
- fixed stack alignment setting on reference FirstEntryMinmal.s

PR #23, opened 2015-03-18
Eliminated warnings for MCBSTM32 solutions (Keil MDK)

cw2:
`Could Not Find xxx.lib`
`Could Not Find xxx.lib.manifest`
`Warning: A1608W: MOV pc,<rn> instruction used, but BX <rn> is preferred`
`Warning:  #177-D: variable "xxx" was declared but never referenced`
`Warning:  #177-D: label "TinyCLR_Cleanup" was declared but never referenced`
`Warning:  #546-D: transfer of control bypasses initialization of...`
`Warning:  #1-D: last line of file ends without a newline`
`Warning:  #69-D: integer conversion resulted in truncation`


smaillet-ms:
Thanks for doing this, some of these have been on my list of annoying things to make go away! :smile: 

Please split this out to separate pull requests, too much is comingled. Especially the targets file updates as I don't think we want solve the problem the way you've done it here. The build should be deleting those files on a clean/rebuild. Silencing that message should be done at the actual source of the message instead. That is,  the build task implementation shouldn't be generating a normal priority message. In effect the task is treating a perfectly normal and anticipated situation as exceptional, which generally leads to confusion. At most it should treat it as a low priority informational message.

cw2:
Ok, I understand. So I'll create a new pull request just for the compiler warnings and preserve the commits. Ok?

PR #24, opened 2015-03-18
Fixed check for CMSIS install (wasn't updated to new CMSIS location)

smaillet-ms:
Fixed InitialTargets to test for ARM.CMSIS.pdsc file in the CMSIS root location (which was recently moved but the checks were still looking in the old location)


mortezag:
:+1:

PR #28, opened 2015-03-20
Eliminate need to manually set FLAVOR_WIN 

smaillet-ms:
- Set default for solution builds to Release
- Fixed build so manually setting FLAVOR_WIN isn't needed anymore
- set default MEMORY type to FLASH as that's more common with smaller micro nowadays
- fixed syntax errors in Microsoft.SPOT.Support.Settings (which seems to be unused...)
- Silenced message about not finding lib and manifest files (the conditions for detecting the existence of files to delete were wrong; This is a minor tweak to the pull request #23 from @cw2)


mortezag:
:+1: 
(Please update Getting Staretd wiki if needed)

PR #31, opened 2015-03-20
Feature: command aliases

smaillet-ms:
* Updated setenv_base to check for an `alias.txt` file in %SPOCLIENT% to apply DOSKEY aliases
* updated gitignore to ignore the alias.txt file as it is optional and per user

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.log
 *_assemblyinfo.txt
 *.feedback
+/alias.txt
 
 # CMSIS and CMSIS_RTX
 /CMSIS/CMSIS/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -29,6 +30,10 @@
 [Oo]bj/
 ipch/
 
+#uVision
+*.uvguix*
+Listings/
```
mortezag: What are these Listings folders? how they are created?

mortezag:
:+1:

smaillet-ms:
The listings folders are created by the uVision tool as an intermediate output folder, so they shouldn't be included in the tree as they are just a side effect of using uVision for deployment over JTAG

PR #32, opened 2015-03-20
UV4 Projects for deployment

smaillet-ms:
Added uV4 projects for deploying build output to an MCBSTM32F400 via uLINK-Pro JTAG probe


mortezag:
:+1:

PR #34, opened 2015-03-21
Add warnings for missing VSSDKs

smaillet-ms:
added warning message for missing VSSDK installations when building VSIX packages


mortezag:
:+1:

PR #36, opened 2015-03-22
Eliminated circular dependency between FLAVOR_WIN and Configuration. Fixes #33

### LINKED ISSUE #33: VSIX generation fails with VSSDK1039: Source.extension.vsixmanifest file not found
I followed building guide but even though building SDK runs successfully, VSIX generation fails with mentioned VSSDK1039 error. How could I verify "correctness" of my VSSDK setup?

Exact (and complete) error reported is:

C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v12.0\VSSDK\Microsoft.VsS
DK.targets(59,5): error VSSDK1039: Source.extension.vsixmanifest file not found
 in project. If a file with this name is present in the project, make sure the
build action is set to "None". [C:\NETMF\netmf-interpreter\Framework\CorDebug\v
s12\cordebugvs12.csproj]

Thanks!

cw2:


cw2:
Also, please review usage of `FLAVOR_WIN` in init.cmd - it is being used in several other variables, even though it [env. var] does not exist anymore.

smaillet-ms:
To fully work with your changes to buildenv.props you;ll need to add the following to init.cmd around line 19: 
if /I "%FLAVOR_WIN%"   == "Debug"   set FLAVOR_WIN=Debug
if /I "%FLAVOR_WIN%"   == "Release" set FLAVOR_WIN=Release
if    "%FLAVOR_WIN%"   == ""        set FLAVOR_WIN=Release

This covers the some of the story, unfortunately it doesn't cover it all. As it will break as soon as you try and build a debug firmware image with /p:FLAVOR=Debug, which is why I removed the setting from init.cmd. The underlying problem is the plethora of variables that represent the classic "Configuration". FLAVOR, FLAVOR_DAT, FLAVOR_WIN,etc... basically we've got more flavors than a Ben and Jerry's :frowning: add to that that the build tries to build EVERYTHING including the tools that build the rest based on the same settings (FLAVOR_WIN governs which version of the tools is built/used )

cw2:
Hm, I am not sure I understand why the content of init.cmd should be reverted (before commit ab03f51e7f3eb17a255720c4c9c3e479a5a7a219)?

I have compared msbuild diag logs and it is like this:

1. Before commit ab03f51e7f3eb17a255720c4c9c3e479a5a7a219 at the beginning of cordebugvs12.csproj build the value of `FLAVOR_WIN = Release` and both `CreateVsixContainer` and `DeployExtension` are set to _false_. Build succeeds.
2. After commit ab03f51e7f3eb17a255720c4c9c3e479a5a7a219  there is no FLAVOR_WIN set in the init.cmd, so its value is empty at the beginning of cordebugvs12.csproj build and `CreateVsixContainer` and `DeployExtension` are set to _true_ causing the build to fail. (see comment in issue #33).
3. After this fix, we are back to 1. The logs are basically identical.

I do not know what exactly happens in case 2, I have not debugged the msbuild session. For some reason, the properties
```
<PropertyGroup Condition="'$(Configuration)' == 'Release'">
    <CreateVsixContainer>False</CreateVsixContainer>
    <DeployExtension>False</DeployExtension>
  </PropertyGroup>
```
are not set, which can obviously happen if `Configuration` does not match. The only explanation I could come up with is that due to circular reference, the Configuration is set to empty FLAVOR_WIN value, i.e. neither Debug nor Release.

The fix does not attempt to solve anything else - if there were some other problems before, they remain...

smaillet-ms:
Turns out this is the tip of the iceberg on several issues. I've got what appears to be a fix of the underlying and related problems. I'm running some test builds now and hope to post a pull request later tonight

smaillet-ms:
these changes and more have been incorporated into #37

Issue #30, opened 2015-03-20
missing vsix


fredrik-ngenic:
Hi,
don't know if this is the right place but...

When i run the build_sdk.cmd I get the sdk msi in the place that is mentioned in the wiki,
but the vsix files aren't where they're supposed to (BuildOutput\public\Release\Server\dll).
And I can't find them anywhere else either...
Any ideas?

Regards 
Fred

smaillet-ms:
This is the perfect place to ask such questions!
Sounds like you don't have the Visual Studio SDK installed. The build system will skip generating the VSIX files if it detects the SDK is not installed. Now that we've cleaned up the SDK build it's probably worth generating an informative message on that. (e.g. if the build is skipping a particular VSIX it should print a message to ley you know it is doing so and why)

mortezag:
Maybe related to #33

fredrik-ngenic:
Thank you,
I'll look into that

PR #37, opened 2015-03-24
Fix for issue #30 and issue #33

smaillet-ms:
- Fixed vs14 vsix package project to use proper dependencies so it can work against VS2015
- fixed location of import to spot.csharp.host.targets file so that it picks up the NetMFBuildVsSdkVersion property to override VSSDK paths and settings
- updated build_sdk.cmd to include a check for VS2012 SDK as that is required to build the templates in the sdk.dirproj phase of the build.
- Fixed vsix package projects to get proper OutDir setting so input and output files come from the correct location.
- updated BuildEnv.props to set a default FLAVOR value
- updated buildenv.props to set configuration based on FLAVOR (unless FLAVOR==RTM in which case Configuration should be Release)
- updated BuildEnv.props to set BUILD_TREE_CLIENT and BUILD_TREE_SERVER unconditionally based on BUILD_TREE, which will be set based on Configuration. Otherwise the BUILD_TREE_XXX and derived paths will always have the value hardcoded into setenv_base/init.cmd as a default
- removed redundant build of MSI and VSIX packages from setup\build.dirproj as build_sdk itself can be used for developer builds of the SDK
- updated BuildEnv.Props to include changes from @cw2 (pull request #36) to handle setting flavor_win

### DIFF
--- a/build_sdk.cmd
+++ b/build_sdk.cmd
@@ -60,5 +60,5 @@ GOTO :EOF
 goto :EOF
 
 :MissingVSSDK
-@ECHO ERROR: Visual Studio 2012 SDK (VSSDK) was not detected, this SDK is required to build the .NET Micro Framework SDK source code
+@ECHO ERROR: Visual Studio 2013 SDK (VSSDK) was not detected, this SDK is required to build the .NET Micro Framework SDK source code
 goto :EOF
--- a/build_sdk.cmd
+++ b/build_sdk.cmd
@@ -9,6 +9,9 @@ if "%1" == "-?" goto :ShowUsage
 if /i "%1" == "/h" goto :ShowUsage
 if /i "%1" == "-h" goto :ShowUsage
 
+if /i "%VSSDK120Install%"=="" goto :MissingVSSDK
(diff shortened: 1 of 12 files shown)


build_sdk.cmd
```diff
@@ -54,3 +57,8 @@ GOTO :EOF
 @ECHO     RELEASE_NAME = Name for the release [ Default = "(%%USERNAME%%)"]
 @echo example:
 @ECHO     build_sdk.cmd 1234 \\NETMFBLD02\Builds\69423 client_vNext "(RC1)"
+goto :EOF
+
+:MissingVSSDK
+@ECHO ERROR: Visual Studio 2012 SDK (VSSDK) was not detected, this SDK is required to build the .NET Micro Framework SDK source code
```
mortezag: shouldn't this be 2013?

mortezag:
:+1:

build_sdk.cmd
smaillet-ms: oops, yes it should say VS2013!

PR #40, opened 2015-03-25
Eliminated build_sdk compiler warnings (partial fix for issue #5)

cw2:
Partial fix for issue #5

Supressed in OpenSSL
`warning C4018: '>' : signed/unsigned mismatch`
`warning C4065: switch statement contains 'default' but no 'case' labels`
`warning C4390: ';' : empty controlled statement found; is this the intent?`
`warning C4996: '<apiname>': was declared deprecated`

Fixed
`warning C4018: '<op>' : signed/unsigned mismatch`
`warning C4244: '+=' : conversion from 'INT64' to 'UINT32', possible loss of data`
`warning C4700: uninitialized local variable '<var>' used`
`warning C4793: '<func>' : function compiled as native`
`warning C4800: 'BOOL' : forcing value to bool 'true' or 'false' (performance warning)`

### DIFF
--- a/DeviceCode/pal/TouchPanel/touchpanel_driver.cpp
+++ b/DeviceCode/pal/TouchPanel/touchpanel_driver.cpp
@@ -325,9 +325,9 @@ TouchPoint* TouchPanel_Driver::AddTouchPoint(UINT16 source, UINT16 x, UINT16 y, 
         if (lastAddedX != 0xFFFF)
         {
             /// dist2 is distance squared.
-            UINT32 dx = abs(x - lastAddedX);
-            UINT32 dy = abs(y - lastAddedY);
-            UINT32 dist2 = dx * dx + dy * dy;
+            INT32 dx = abs(x - lastAddedX);
+            INT32 dy = abs(y - lastAddedY);
+            INT32 dist2 = dx * dx + dy * dy;
 
             /// Ignore this point if it is too far from last point.
             if (dist2 > g_TouchPanel_Sampling_Settings.MaxFilterDistance) return NULL; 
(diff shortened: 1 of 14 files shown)


DeviceCode/pal/TouchPanel/touchpanel_driver.cpp
```diff
@@ -460,8 +460,11 @@ HRESULT TouchPanel_Driver::GetSetTouchInfo(UINT32 flags, INT32* param1, INT32* p
             // Touch Move events are queued up and sent at the given time frequency (StylusMoveFrequency)
             // We should not set the move frequency to be less than the sample frequency, otherwise there
             // would be no data available occassionally.
-            else if(ticks < min_ms_per_touchsample) return CLR_E_OUT_OF_RANGE;
-            else                                    ticks = TOUCH_PANEL_SAMPLE_MS_TO_TICKS(ms_per_touchmove_event);
+            else
```
smaillet-ms: This appears to be a change in logic not a warning fix. Can you explain what the problem was and how this fixes it?

Solutions/Windows2/TinyCLR/Various.cpp
```diff
@@ -437,6 +437,8 @@ void HAL_Windows_FastSleep( INT64 ticks )
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
+#pragma unmanaged
```
smaillet-ms: prefer using #pragma managed(push, off) + #pragma managed(pop) instead of directly forcing it to managed. This allows the code to work correctly even if built in a purely unmanaged configuration.

DeviceCode/pal/TouchPanel/touchpanel_driver.cpp
cw2: The original code uses uninitialized variable `ticks` in the condition on line 463:
```
INT32 ticks;
...
if(ms_per_touchmove_event == 0) ticks = 0x7FFFFFFF;
else if(ticks < min_ms_per_touchsample) return CLR_E_OUT_OF_RANGE; // <------ ticks is not initialized
else ticks = TOUCH_PANEL_SAMPLE_MS_TO_TICKS(ms_per_touchmove_event);
```
The fix moves range check after the `ticks` value is set to TOUCH_PANEL_SAMPLE... , which is most likely the original intent (?)

Solutions/Windows2/TinyCLR/Various.cpp
cw2: Good point, thanks. Fixed in 511cc680de3d7e565d25a2efb2035419f679915a.

PR #42, opened 2015-03-25
Fix Deploy failures on MCBSTM32F400

smaillet-ms:
- Updated Debug engine to provide useful messages on flash erase/write errors during deployment.
- Added Debug engine trace event for erase memory
- Fixed MCBSTM32F400 Blck config for the external FLASH to use the whole flash for Deployment, which is waht the scatterfile.xml is set up to assume. (this caused the device to NACK the sector erase which was reported with little more detail than E_FAIL in the IDE :( )
- Fixed the ArmLinkerOffset bug for the LoadRegion to reflect current imperically determined offset value. (Awaiting full fix from ARM linker team to eliminate this hack)
- added *.user to the .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@
 *.suo
 *.vspscc
 *.vssscc
+*.user
 [Bb]in/
 [Oo]bj/
 ipch/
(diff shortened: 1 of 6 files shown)


mortezag:
:+1:

PR #43, opened 2015-03-26
Fix uVision project settings for TinyCLR download

smaillet-ms:
Set download options to skip verification stage as the IDE will reset the hardware after flash and can't verify the data without reprogramming the external memory controller again.


mortezag:
:+1:

PR #44, opened 2015-03-26
Add Initialization and TinyHal libraries for STM32F4

mortezag:


DeviceCode/Targets/Native/STM32F4/DeviceCode/Initialization/tinyhal.cpp
```diff
@@ -0,0 +1,752 @@
...
+#include <tinyhal.h>
+#include "stm32f4xx.h"
+
+#if defined(PLATFORM_ARM_OS_PORT)
+#include <cmsis_os.h>
+#include <cstdint>
+#endif
+
+// these define the region to zero initialize
+extern UINT32 Image$$ER_RAM_RW$$ZI$$Base;
```
smaillet-ms: The Linker vairables Image$$... and Load$$... are specific to the ARM compiler and not generally useable on other tool chains. (i.e. gcc) furthermore we know there's a bug in the linker where it gets the Load$$ value stored in the literal pool wrong for this. Can we pull all of this this out to a toolset specific file at least?

DeviceCode/Targets/Native/STM32F4/DeviceCode/Initialization/tinyhal.cpp
```diff
@@ -0,0 +1,752 @@
...
+
+    // have to initialize the blockstorage first, as the USB device needs to update the configure block
+
+    BlockStorageList::Initialize();
+
+    BlockStorage_AddDevices();
+
+    BlockStorageList::InitializeDevices();
+
+    //FS_Initialize();
```
smaillet-ms: These shouldn't be commented out here in a general solution as they will never be called, even if the related features are added to the TinyCLR.proj. Ultimately we need a better way to leave these out when not needed so they don't force the inclusion of other dependencies and stubs, etc...

DeviceCode/Targets/Native/STM32F4/DeviceCode/Initialization/tinyhal.cpp
```diff
@@ -0,0 +1,752 @@
...
+// For reasons unknown, the ARM linker is getting the
+// fixup for the LoadRegion symbol incorrect. The data
+// is located correctly but the fixed up pointer stored
+// in the literal pool that this code loads for the address
+// of the load base (src) is off by some factor. In initial
+// testing it was always 0x90, unfortunately it turns out
+// not to be consistent and bumped up to 0xED0, and is now
+// back at 0x90... Sigh... Hope to hear back from ARM support
+// on this soon.
+const UINT32 ArmLinkerLoadRegionOffsetHack = 0x00000090;
```
smaillet-ms: This hack can't go here. It is completely dependent on the entire set of code and data for the full firmware image, and therefore VERY solution specific.

smaillet-ms:
It might be helpful if you could add an explanation of why this is necessary to the description. (Generally a good ide/practice for something without a specific bug or feature issue)

mortezag:
I'll create another pull request.

PR #47, opened 2015-03-26
MCBSTM32F400: Refactor Initialization and TinyHal libraries

mortezag:
Update Initialization and TinyHal libraries of MCBSTM32F400 so that they can be used in SolutionWizard


smaillet-ms:
:+1:

PR #49, opened 2015-03-27
Threadingsample

ColinA-MSFT:
Port and update Clock, CustomUI, and Threading code samples from CodePlex.


Product/Samples/Clock/Clock/GPIOButtonInputProvider.cs
```diff
@@ -0,0 +1,175 @@
...
+            // get back an InputProviderSite object which forwards the input 
+            // report to the input manager, which then places the input in the 
+            // staging area.
+            site = InputManager.CurrentInputManager.RegisterInputProvider(this);
+
+            // Create a delegate that refers to the InputProviderSite object's 
+            // ReportInput method.
+            callback = new DispatcherOperationCallback(delegate(object report)
+                {
+#if MF_FRAMEWORK_VERSION_V3_0
```
smaillet-ms: Let's lose the legacy conditional code. v3.0 was almost 8 years ago...

smaillet-ms:
The file system sample seems more like a complex UI sample than something demonstrating the FS support. What FS support was there is lost in all the extra UI. We should probably rename this sample to reflect it's actual coverage and create a new FS sample that is focused on just the FS without all the UI pizazz. (Simple Debug print is good enough)

Product/Samples/Clock/Clock/GPIOButtonInputProvider.cs
ColinA-MSFT: This is all generated.  I've updated this instance, but we should make the same changes to the template.

smaillet-ms: ok, go ahead and file a new bug to fix that.

Product/Samples/Clock/Clock/Program.cs
```diff
@@ -0,0 +1,134 @@
...
+        void TimeService_SystemTimeChanged(object sender, SystemTimeChangedEventArgs evt)
+        {
+            Debug.Print("System Time Changed.");
+        }
+
+        private void OnButtonUp(object sender, RoutedEventArgs evt)
+        {
+            ButtonEventArgs e = (ButtonEventArgs)evt;
+
+            int timeZoneOffset = -8 * 60;
```
mortezag: I think it better to make this value a const or at least put a comment on top of it.

mortezag:
The header documentations are not consistent, some of methods have them but some not. I am not sure if it valuable to put hear documentation on every method but I just wanted to bring it up.

mortezag:
In general, I believe the samples are long. They are more like a show case. If the goal is to demo something interesting they are good, but if the goal is to just show how to do a specific thing they are too complicated.

mortezag:
:+1:

Product/Samples/Clock/Clock/Program.cs
ColinA-MSFT: I'll rename it and make it const so its purpose is clearer.

PR #50, opened 2015-03-28
Fix collisions in GUIDs of libraries

mortezag:
Each library should have a unique GUID, but some libraries have same GUIDs (probably due to manual modifications of dotNetMF.proj files). This commit fixes a few collisions between library GUIDs.


smaillet-ms:
:+1:

PR #52, opened 2015-03-28
Move reducesize property before the import of Microsoft.SPOT.System.Settings

mortezag:
The main point of this change is to enable conditional inclusion of settings in a solution settings file based on reducesize property (for example, include CMSIS_RTOS settings for TinyClr but not for TinyBooter)

### DIFF
--- a/CLR/Tools/PlatformDesigner/ComponentObjectModel/MsBuildWrapper.cs
+++ b/CLR/Tools/PlatformDesigner/ComponentObjectModel/MsBuildWrapper.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Collections;
 using System.Text;
 using System.IO;
+using System.Linq;
 using XsdInventoryFormatObject;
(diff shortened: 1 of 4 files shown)


CLR/Tools/PlatformDesigner/ComponentObjectModel/MsBuildWrapper.cs
```diff
@@ -3137,6 +3138,13 @@ internal void SaveProjectProj(MFProject mfproj)
 
                     ProjectPropertyGroupElement bpg = SaveStringProps(proj, mfproj, tbl);
 
+                    if (mfproj.Properties.Any(p => p.Name == "reducesize"))
```
smaillet-ms: This appears to be enumerating the properties twice. Any reason not to just use FirstOrDefault() to get both cases with only one pass?

mortezag:
@smaillet-ms would you have a look at the updated changes.

smaillet-ms:
:+1:

PR #57, opened 2015-03-31
Remove unnecessary managed code crash from TristatePort

Sukasa:
There's no need to artificially increase the learning curve by throwing a managed code exception here.  If .Active is being set to its current value, the existing code will avoid changing the port without the need for a crash.


smaillet-ms:
In general it is best to submit a bug before posting  a pull request with code changes so the issue can be reviewed and discussed. For small changes we can skip that but we need to have an explanation of what the problem really is, how the fix resolves it, and how the change in behavior wont' negatively impact other parts of the system.

Sukasa:
My apologies for not following the existing convention.  I've created an issue for this to follow up on that.

mortezag:
:+1:

ColinA-MSFT:
:+1:

smaillet-ms:
In order to provide contributions to the .NET Micro Framework please fill out and submit a [Contribution License Agreement (CLA)](https://github.com/NETMF/netmf-interpreter/wiki/Contribution%20License%20Agreement). This is needed to protect everyone and ensure that the community can legitimately use your contributions as part of the .NET Micro Framework

Sukasa:
I've filled out and signed the CLA, as requested.

PR #59, opened 2015-03-31
Import LCD_HAL libcatproj from TinyCRT proj

mortezag:
LCD_WriteFormattedChar is referenced from tinycrt.cpp. This fix makes sure that at least the stub for LCD gets included during the creation of a solution using SolutionWizard.


smaillet-ms:
:+1:

PR #60, opened 2015-03-31
SolutionWizard: Fix IsBootloaderProject method

mortezag:
Change the logic of IsBootloaderProject method to not only check reduceSize property but also make sure that its value is true


smaillet-ms:
:+1:

PR #68, opened 2015-04-02
Remove PKCS11_Storage_PAL libcat form Crypto featureproj

mortezag:
It seems that it is not actually required!


smaillet-ms:
:+1:

PR #69, opened 2015-04-02
Add Backlight and Piezo libcats to SPOT lib

mortezag:
As these Backlight and Piezo are referenced from SPOT lib


smaillet-ms:
:+1:

PR #71, opened 2015-04-03
re-enabled support for ETM tracing for profiling

smaillet-ms:
- re-worked scatterfile XML for TinyCLR to allow chip pins to be used as trace pins in debug builds
- Added ifdef DEBUG checks in GPIO initialization for SRAM and FLASH to exclude ETM pins in debug mode so they are available for profiling
- removed old forced includes to the RTIP stack from MDK targets file
- disabled round-robin scheduling in RTX config as it isn't required and produces functionality much closer to what the previous CLR had. The CLR is the only continuosly running thread everything else is event driven and runs when the CLR is has nothing elas eto do (when it otherwise would have run the stack code as a continuation anyway)


mortezag:
:+1:

PR #72, opened 2015-04-06
Reduce time to deploy to device by skipping unnecessary block erasures.

ColinA-MSFT:
This fix should also help to improve hardware longevity by removing unnecessary flash writes.

### DIFF
--- a/CLR/Debugger/Debugger.cpp
+++ b/CLR/Debugger/Debugger.cpp
@@ -622,7 +622,7 @@ bool CLR_DBG_Debugger::AccessMemory( CLR_UINT32 location, UINT32 lengthInBytes, 
                 {
                     case AccessMemory_Check:
                     case AccessMemory_Read:
-                        if (deviceInfo->Attribute.SupportsXIP)
+                        if(deviceInfo->Attribute.SupportsXIP)
                         {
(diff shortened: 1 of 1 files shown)


CLR/Debugger/Debugger.cpp
```diff
@@ -622,7 +622,7 @@ bool CLR_DBG_Debugger::AccessMemory( CLR_UINT32 location, UINT32 lengthInBytes,
                 {
                     case AccessMemory_Check:
                     case AccessMemory_Read:
-                        if(deviceInfo->Attribute.SupportsXIP)
+                        if (deviceInfo->Attribute.SupportsXIP)
```
smaillet-ms: Please don't add whitespace there like that
The original was fine as it was, and the preferred style

CLR/Debugger/Debugger.cpp
```diff
@@ -650,25 +650,34 @@ bool CLR_DBG_Debugger::AccessMemory( CLR_UINT32 location, UINT32 lengthInBytes,
...
+                        }
                         break;
+
                      default:
                         break;
                 }
 
 
-                if(!success)
+                if (!success)
```
smaillet-ms: Restore this.

smaillet-ms:
:+1:

mortezag:
:+1:

PR #75, opened 2015-04-10
Fix usb config

smaillet-ms:
- Cleaned up entry point assembly code in prep for supporting CMSIS-CORE based startup (when full CMSIS-CORE is in place all of this initialization stuff goes way)
- Fixed bug in MFCongiHelper for MFDeploy to allow setting device name correctly
- Moved MCBSTM32F400 configuration into external NOR flash to make more room for code and to ensure that the configuration region is accessible from MFDEPLOY (device libraries for the wire protocol assume only one block storage device is available and therefore the configuration region must be in the same device as the deployment )
- Updated scatter file XML so that it's easier to compare and contrast boot loader layout with full CLR as there really isn't much that is or should be different between them. Consistency helps eliminate errors (like the boot loader thinks the configuration region is someplace different from the CLR...)
- removed tabs from a number of files


mortezag:
:+1:

PR #77, opened 2015-04-14
MFTestSystem: Set the exit code based on the test results

mortezag:


Test/Platform/Tools/MFTestSystem/TestSystem.cs
```diff
@@ -61,12 +61,12 @@ static void Main(string[] args)
                 if (args[0] == "/?")
                 {
                     DisplayHelp();
-                    return;
+                    return -1;
```
smaillet-ms: To enable more useful troubleshooting, lets define an enumeration with a unique identifier for each error case. Furthermore we should generate a text error message on Console.Error as well.

Test/Platform/Tools/MFTestSystem/TestSystem.cs
```diff
@@ -61,12 +61,12 @@ static void Main(string[] args)
...
                 {
                     DisplayHelp();
-                    return;
+                    return -1;
                 }
                 else if (String.Equals(args[0], "-help", StringComparison.InvariantCultureIgnoreCase))
                 {
                     DisplayHelp();
-                    return;
+                    return -1;
```
smaillet-ms: Displaying help should not produce an error code. (This can mess up CMD and bat scripts that set ERRORLEVEL based on the return code of the last run process)

Test/Platform/Tools/MFTestSystem/TestSystem.cs
```diff
@@ -79,19 +79,20 @@ static void Main(string[] args)
...
                     "An uncaught exception was thrown when invoking RunTests: {0}",ex.ToString()));
+                return -1;
             }
         }
 
         /// <summary>
         /// Use this method to execute tests.
         /// </summary>
-        public void RunTests()
+        public bool RunTests()
```
ColinA-MSFT: Avoid using bool for success or failure.  Use error codes or exceptions.  In this case, we should throw. Main can catch and exit the process with an error code.

mortezag:
@smaillet-ms @ColinA-MSFT  updated

smaillet-ms:
:+1:

ColinA-MSFT:
:+1:

PR #80, opened 2015-04-17
Improve device deployment messages

smaillet-ms:
Added DeviceName to the deployment message when trying to connect to a device (e.g. `"[<device name>] Iteration xx..."`) This helps notify the user of which device the project was last set to use, which may not be the device currently connected! This helps in diagnosing any connection problems due to simply forgetting to select the device or accidentally picking the wrong one, etc...


Framework/Debugger/UsbDevicePortNameEqualityComparer.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace Microsoft.SPOT.Debugger
 {
-    /// <summary>Equality comparer to compare devce names ofr USB ports independent of the interface</summary>
+    /// <summary>Equality comparer to compare devce names of USB ports independent of the interface</summary>
```
brianingenito: While in here you might as well fix the typo on device too.

smaillet-ms: Doh! Note to self - adjust caffeine dose before coding. :coffee:

PR #85, opened 2015-04-24
Corrected parity and stop bits setup.

bheeb:


DeviceCode/Targets/Native/STM32/DeviceCode/STM32_USART/STM32_usart_functions.cpp
```diff
@@ -91,13 +91,14 @@ BOOL CPU_USART_Initialize( int ComPortNum, int BaudRate, int Parity, int DataBit
     
     // control
     UINT16 ctrl = USART_CR1_TE | USART_CR1_RE;
-    if (DataBits == 9) ctrl |= USART_CR1_M;
-    if (Parity) ctrl |= USART_CR1_PCE;
+    if (Parity) { ctrl |= USART_CR1_PCE; DataBits++; }
```
smaillet-ms: Please split this to two lines. Many embedded debuggers can't set breakpoints on sub line spans so you can't break only when the condition is true.

PR #86, opened 2015-04-24
Corrected parity bit setup.

bheeb:


DeviceCode/Targets/Native/STM32F4/DeviceCode/STM32F4_USART/STM32F4_usart_functions.cpp
```diff
@@ -150,7 +150,7 @@ BOOL CPU_USART_Initialize( int ComPortNum, int BaudRate, int Parity, int DataBit
     // control
     UINT16 ctrl = USART_CR1_TE | USART_CR1_RE;
     if (Parity)
-        ctrl |= USART_CR1_PCE; DataBits++;
+        {ctrl |= USART_CR1_PCE; DataBits++;}
```
smaillet-ms: Please align the { with the if and split the statements onto individual lines. Some debuggers can't set breakpoints or step on statements on a sub line basis.

PR #88, opened 2015-05-04
Core and MCBSM32F400 updates from Alljoyn demo work

smaillet-ms:
This does not include Alljoyn (as there's more work to do on that), but does include the changes to the core and MCBSTM32F400 platform that are applicable beyond Alljoyn. 
- Corrected ConfigHelper.cpp to test for the constant sizeof() value embedded as a member of the config section to validate the structure is good instead of testing agains 0xFFFFFFFF, which works only on erased flash. This prevents bogus data that remains in flash from being treated as a valid config if the config section is moved to a new location during development, and matches what the desktop side uses for validation.
- Fixed how the STM32F400 Ethernet driver posts packets to the network stack. This is now done by posting directly to the stack from the ISR, instead of using a continuation, whose sole job was to dispatch the work to the network stack. (e.g. what we had was effectively the isr post to the CLR thread which then just posted whatever it got onto the TCPIP thread. As it turns out LWIP has a built in function to support this scenario without the continuation as an intermediate step.
- Updated diagnostic messages for connectivity to help make clear what device is used when trying to connect.
- fixed bug in Block storage map for the external FLASH on the MCBSTM32F400. It left a gap of 1 unallocated sectiion
- Increased OS_STKSIZE for MCBSTM32F400 as it was really close to the edge (and our Alljoyn demo hit the limit)
- updated Linker bug hack in MCBSTM32F400 to 0 as workaround from ARM seems to function. (leaving code and comments in place in case it goes pear shaped agai
- modified global new/delete for MCBSTM32F400 to use private_xxx heap APIs. (BitConverter class uses new/delete, this works around the issue as we have a private heap in this solution but isn't a long term answer to this)
- reset MCBSTM32F400 to use proper OS threading aware GlobalLock support


Solutions/MCBSTM32F400/DeviceCode/Initialization/tinyhal.cpp
```diff
@@ -407,8 +407,15 @@ void HAL_UnReserveAllGpios()
...
-const UINT32 ArmLinkerLoadRegionOffsetHack = 0x00000090;
+// on this soon. ARM has confirmed the bug as related to the
+// use of a non-compressed region in the scatter file. Since
+// the code here doesn't understand the linker compression
+// the region is uncompressed, however the linker bug is 
+// that an uncompressed region that follows after a compressed
+// one will get the invalid Load$$xxx$$Base value. If the
+// scatter file is updated to ensure all regions occuring
+// before this one are also uncompressed it works ok.
+const UINT32 ArmLinkerLoadRegionOffsetHack = 0x00000000;
```
mortezag: Is this Linker problem fixed forever? or this just happened to be zero for now?

mortezag:
:+1:

Solutions/MCBSTM32F400/DeviceCode/Initialization/tinyhal.cpp
smaillet-ms: Hopefully, yes. The workaround to make sure all sections before the LWIP one are uncompressed seems to work. The "real" fix will come with an update to the ARM tool chain. Once that happens we'll remove the hack.

Issue #13, opened 2015-03-13
MCBSTM23F400 Solution does not support newer rev1.2 boards with updated PHY


smaillet-ms:
The MCBSTM32F400 rev1.2 boards use a KSZ8081 PHY instead of the ST802RTI. Ethernet/networking support therefore does not work on the newer board revisions.

smaillet-ms:
Fixed in #90

PR #90, opened 2015-05-05
MCBSTM32F400 v1.2 support ( Issue #13 )

smaillet-ms:
Fix for issue #13 
Updated Ethernet driver to support the new PHY used on MCBSTM32F400 v1.2 boards This now supports both revisions of the board.


mortezag:
:+1:

PR #91, opened 2015-05-05
Fix Arrays tests solution file

mortezag:
RunTests tool was unable to run the Arrays test solution. I updated the sln and csproj file to be similar to other test solutions. Now, RunTests tool can build and run this test solution.


smaillet-ms:
:+1:

PR #93, opened 2015-05-06
correct path for lwip_1_4_1_os

josesimoes:
There was a typo in the path in <HFiles Include...>. Was pointing to
$(SPOCLIENT)\devicecode\psl\ instead of $(SPOCLIENT)\devicecode\pal\.

Wondering: if this was compiling on your end are these includes really necessary?...


smaillet-ms:
Nice catch, thanks.
The HFiles element is a bit of an odd thing. It comes from the fairly lengthy history of the build system , which predates VS support for VCXPROJ files. What those do is specify that the build should look at those header files when doing dependency checking to determine if a rebuild is needed. So, they aren't technically required, but without them you can spend a lot of time scratching your head or pulling your hair out trying to figure out why a change you made to a header file doesn't seem to have any impact when you build. Forcing a rebuild won't see this as everything is always built but the build takes longer as a result.

PR #94, opened 2015-05-06
Fix tests regarding interface re-implementation

mortezag:
A fix for the tests which exercises Interface re-implementation rules (section 13.4.6 of C# Language Specification)

(This is not a fix for the actual problem in CLR but just the fix for its tests)


smaillet-ms:
:+1:

PR #97, opened 2015-05-08
Add Windows.Devices assembly

ColinA-MSFT:
Add a new assembly for Windows.Devices, with initial support for Gpio and I2c.


Framework/Core/Windows/Devices/Gpio/GpioPinEventListener.cs
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        // IEventProcessor methods
+
+        public BaseEvent ProcessEvent(uint data1, uint data2, DateTime time)
+        {
+            return new GpioPinEvent
+            {
+                // Data1 is packed by PostManagedEvent, so we need to unpack the high word.
+                PinNumber = (int)((data1 >> 16) & 65535u),
```
mortezag: What is the purpose of the '&' operation? Don't (data1 >> 16) and (data >> 16) & 65535u always have the same value here?

Framework/Core/Windows/Devices/Enumeration/DeviceInformation.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        // Private methods
+
+        private static DeviceInformation[] FindAllI2c()
+        {
+            // TODO: Determine whether this bus exists.
+            var devices = new DeviceInformation[1];
+            devices[0] = new DeviceInformation("I2C0", true);
+            return devices;
+        }
```
mortezag: I do not get the logic of this method. What this method is supposed to return? All I2C devices/Ports? Why does it always return the same array?

Framework/Core/Windows/Devices/Enumeration/DeviceInformation.cs
```diff
@@ -0,0 +1,98 @@
...
+        // We don't support full AQS in the Micro Framework. Instead, we use a pre-set list of
+        // hard-coded strings. These strings should be considered opaque, so developers should use
+        // the GetDeviceSelector helpers to ensure future compatibility.
+        internal static string s_I2cPrefix = "I2C";
+        internal static string s_SpiPrefix = "SPI";
+
+        // Private fields
+
+        private string m_id;
+        private bool m_isDefault;
```
mortezag: I am not sure if we have any C# coding conventions or not. But I noticed that some of the conventions used here are different from other codes; for example, fields at the end of a class, or C style prefix for members name. Maybe we should use StyleCop in future. For now, I just wanted to bring it up.

Framework/Core/Windows/Devices/Gpio/GpioPinEventListener.cs
```diff
@@ -0,0 +1,89 @@
...
+            // Avoid calling this under a lock to prevent a potential lock inversion.
+            if (pin != null)
+            {
+                pin.OnPinChangedInternal(pinEvent.Edge);
+            }
+
+            return true;
+        }
+
+        // Public methods
```
mortezag: I believe #region directive is more suitable for grouping class members.

Framework/Core/Windows/Devices/I2c/I2cDevice.cs
```diff
@@ -0,0 +1,206 @@
...
+        public uint BytesTransferred;
+    }
+
+    public sealed class I2cDevice : IDisposable
+    {
+        // Construction and destruction
+
+        internal I2cDevice(ushort slaveAddress, I2cBusSpeed busSpeed)
+        {
+            int clockRateKhz = 100;
```
mortezag: Please make any hardcoded value a constant with a self-explanatory name or some comments.

Framework/Core/Windows/Devices/I2c/I2cEnums.cs
```diff
@@ -0,0 +1,21 @@
+namespace Windows.Devices.I2c
+{
+    public enum I2cBusSpeed
+    {
+        StandardMode = 0,   // 100kHz
```
mortezag: I am just curious to know why you always set the integer value of the first member of an enum to zero. Isn't zero always the default integer value for the first member of any enum?

Framework/Core/Windows/Devices/Enumeration/DeviceInformation.cs
ColinA-MSFT: This is for parity in the API surface of UWP.  Additionally, we will more than likely support more than one I2C bus in the near future.  This lets us do so without changing the API.

Framework/Core/Windows/Devices/Enumeration/DeviceInformation.cs
ColinA-MSFT: We're not especially consistent about coding standards in NetMF.  StyleCop defines the core of it.  Where it leaves ambiguity, I fall back on the general Microsoft-wide coding standards.

In this case, the rule is constructors, then most public to most private.  By convention within each visibility, .NET code bases often use the following order: operators, events, properties, then methods.  Nobody seems to agree on where fields go, though it's either first or last, depending on the code base.

Framework/Core/Windows/Devices/I2c/I2cDevice.cs
ColinA-MSFT: The entire purpose for this variable is to give these values a name (clockRate) and a well-defined unit of measurement (khz).

Framework/Core/Windows/Devices/I2c/I2cEnums.cs
ColinA-MSFT: Microsoft coding standards (including in UWP APIs, which we're mirroring).  The first value should be explicitly assigned, with all following values implicit when possible.

Framework/Core/Windows/Devices/Enumeration/DeviceInformation.cs
mortezag: Actually, StyleCop says "All methods must be placed after all fields" for C#. In any case, also check with @smaillet-ms because I think he was working on a NETMF coding standards document.

mortezag:
Another thing that I should bring up is tests. Do we have any plan to develop some tests for these apis?

Framework/Core/Windows/Devices/Gpio/GpioPinEventListener.cs
ColinA-MSFT: Updated to use region/endregion.  Pragma region no longer appears to work.

ColinA-MSFT:
I have some local tests on my machine, but will need to modify them to fit into our test suite.  The plan is to match existing tests for the equivalent SPOT APIs to ensure similar behavior.

Framework/Core/Windows/Devices/Enumeration/DeviceInformation.cs
smaillet-ms: This is a case I completely disagree with Stylecop and turn off that setting. Reading from top to bottom should be public protected, internal and private. That places the most likely needed by someone using the class up front and the information only needed to someone maintaining the internals of the thing at the bottom.

On the C style variables, loose all the s_, g_, m_ etc... we don't want that in C# code, it's a old legacy C/C++ thing. Since we are now part of the .NET Foundation and closely related to CoreCLR and CoreFX we should probably move towards unifying our style to the one defined there rather than trying to go our own way. (We've got more important things to do than debate style rules)

Framework/Core/Windows/Devices/Gpio/GpioController.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        public bool TryOpenPin(int pinNumber, GpioSharingMode sharingMode, out GpioPin pin, out GpioOpenStatus openStatus)
+        {
+            pin = null;
+
+            try
+            {
+                pin = OpenPin(pinNumber, sharingMode);
+            }
+            catch
```
smaillet-ms: Don't catch all exceptions here. Create an overlod (probably private) to OpenPin that has a flag on whether it should throw exceptions or not. Tryxxx functions don't guarantee no exceptions can occur, they only guarantee that any exceptions known to the implementation won't be thrown. There are still critical and runtime asynchronous exceptions that can occur at any point (i.e. ThreadAbortException) etc... Catching all exceptions like this looses those and alters expected behavior of the system.

CLR/Libraries/Windows_Devices/windows_devices_native_Windows_Devices_Gpio_GpioPin.cpp
```diff
@@ -0,0 +1,234 @@
...
+    CLR_RT_HeapBlock* pThis = stack.This(); FAULT_ON_NULL(pThis);
+
+    if (pThis[ FIELD__m_disposed ].NumericByRef().s1 != 0)
+    {
+        TINYCLR_SET_AND_LEAVE(CLR_E_OBJECT_DISPOSED);
+    }
+
+    value = (CLR_INT64_TEMP_CAST) stack.Arg1().NumericByRef().s8 / TIME_CONVERSION__TO_MILLISECONDS;
+
+    // TODO: Set debounce individually per pin once the HAL has that capability.
```
smaillet-ms: What's the plan to get this implemented? As a general rule code being checked in should not contain "todo:" or any other similar variant of comment indicating known work that isn't completed. Unless they explicitly list an item id for the issue in the issue tracking system governing the code. Consider/review/Future type comments that hint at alternate approaches or capabilities in a future release are OK.

CLR/Libraries/Windows_Devices/windows_devices_native_Windows_Devices_Gpio_GpioPin.cpp
```diff
@@ -0,0 +1,234 @@
...
+{
+    TINYCLR_HEADER();
+
+    GPIO_PIN portId;
+    CLR_RT_HeapBlock* pThis = stack.This(); FAULT_ON_NULL(pThis);
+
+    portId = pThis[ FIELD__m_pinNumber ].NumericByRefConst().u4;
+
+    // Set pin to input state so the pin is in "weak" state and does not drain power. Then release the pin.
+    ::CPU_GPIO_EnableInputPin2( portId, false, NULL, NULL, GPIO_INT_NONE, RESISTOR_PULLUP );
```
smaillet-ms: When we update the HAL to support per pin de-bounce, let's also update it to include an explicit DisablePin() function that way it is semantically clearer on what this is doing and allows the HAL implementation to do whatever is appropriate as the default here using Input mode with a pull-up, may not actually be the best way to do that on a given piece of hardware.

CLR/Libraries/Windows_Devices/windows_devices_stub.cpp
```diff
@@ -0,0 +1,61 @@
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#include "windows_devices.h"
+
+HRESULT Library_windows_devices_native_Windows_Devices_Gpio_GpioController::get_PinCount___I4( CLR_RT_StackFrame& stack )
+{
+    TINYCLR_HEADER();
+    TINYCLR_SET_AND_LEAVE(stack.NotImplementedStub());
```
smaillet-ms: Change the exception return here (and all the others) to Not Supported, Not implemented is mostly a runtime form of TODO: comment, that shouldn't be left in production code. Also, not supported is semantically more correct with a stub library which is literally in place because the given functionality isn't supported in this firmware build.

Framework/Core/Windows/Devices/Enumeration/DeviceInformation.cs
```diff
@@ -0,0 +1,98 @@
+using System;
+
+namespace Windows.Devices.Enumeration
+{
+    public sealed class DeviceInformation
+    {
+        // Construction and destruction
```
smaillet-ms: Loose the blatantly obvious comments such as "constructor" etc... they don't actually provide any information that anyone knowing the language couldn't determine on their own with ease. Do, however, add Full XML doc comments on all public and protected language elements. While we don't have that for all our existing code, we should endeavor to ensure all new libraries include it.

Framework/Core/Windows/Devices/Enumeration/DeviceInformation.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        // FUTURE: This should return DeviceInformationCollection
+        public static DeviceInformation[] FindAll()
+        {
+            return FindAll("");
+        }
+
+        public static DeviceInformation[] FindAll(string aqsFilter)
+        {
+            if (aqsFilter.Equals(""))
```
smaillet-ms: use string.Empty instead of ""

Framework/Core/Windows/Devices/Enumeration/DeviceInformation.cs
```diff
@@ -0,0 +1,98 @@
...
+            {
+                return FindAllI2c();
+            }
+            else if (aqsFilter.Equals(s_SpiPrefix))
+            {
+                return FindAllSpi();
+            }
+
+            // Custom filter; just pass it through as the ID.
+            // TODO: Validate that this bus exists.
```
smaillet-ms: Is there a bug or issue item to track this TODO:? Otherwise fix the issue before checking in.

Framework/Core/Windows/Devices/I2c/I2cDevice.cs
```diff
@@ -0,0 +1,206 @@
...
+        public static I2cDevice FromId(string deviceId, I2cConnectionSettings settings)
+        {
+            if ((settings.SlaveAddress > 0) || (settings.SlaveAddress < 127))
+            {
+                throw new ArgumentOutOfRangeException();
+            }
+
+            if (deviceId != "I2C0")
+            {
+                // TODO: Is this the right exception?
```
smaillet-ms: Make a decision on the exception and go with it, don't leave this commented out as it just ignores this known bad condition.

Framework/Core/Windows/Devices/I2c/I2cDevice.cs
```diff
@@ -0,0 +1,206 @@
...
+            foreach (var transaction in transactions)
+            {
+                bytesRequested += (uint)transaction.Buffer.Length;
+            }
+
+            I2cTransferResult result;
+
+            lock (s_deviceLock)
+            {
+                // TODO: How long should our transaction timeout be? 15ms is the max in UWP, but too short for us.
```
smaillet-ms: Why is it too short? Either way the hardcoded value should be converted to a named const (NOTE: the C# language and .NET runtime require that any reference to the const be replaced with the actual literal value inline in the IL code and therefore the const declaration has no runtime overhead compared to hard coding the value everywhere.)

Framework/Core/Windows/Devices/I2c/I2cDevice.cs
```diff
@@ -0,0 +1,206 @@
...
+                s_device.Config = m_configuration;
+                result.BytesTransferred = (uint)s_device.Execute(transactions, 1000);
+            }
+
+            // TODO: We need a more reliable way to report SlaveAddressNotAcknowledged.
```
smaillet-ms: The HAL/PAL layer should report an error when a transaction is NAK'd. What's the defined UWP behavior for this?

Framework/Core/Windows/Devices/I2c/I2cEnums.cs
smaillet-ms: It is a good idea to be explicit with the 0 value as enums are value types and the default integral value is 0, thus the result of default( T ) where T is an enum is well known and explicitly declared.

Framework/Core/Windows/Devices/Enumeration/DeviceInformation.cs
ColinA-MSFT: I don't have any particularly strong opinions here on what standard we stick to, just so long as we have one.  Ideally, we want it automatically checked by a tool so we don't have to think or argue about it.  For now, that's StyleCop.  We can discuss custom rules for StyleCop in vNext.

Scope prefixes (m_, s_, g_, c_) are in use by CoreCLR and CoreFX, so the code as written matches their conventions.  This is also a Microsoft-wide standard, and has *not* been deprecated.

You may be referring to Hungarian prefixes (p, cb, sz, etc.) which have been deprecated and should not be used in new code; these do appear in the interop layer to match existing components, but we should eliminate this practice in vNext.

Framework/Core/Windows/Devices/Gpio/GpioController.cs
ColinA-MSFT: Agreed.  I was on the fence about this, but you've tipped me over to adding the overload.

CLR/Libraries/Windows_Devices/windows_devices_native_Windows_Devices_Gpio_GpioPin.cpp
ColinA-MSFT: I'll add a new issue to track this and reference in the comment.

CLR/Libraries/Windows_Devices/windows_devices_native_Windows_Devices_Gpio_GpioPin.cpp
ColinA-MSFT: :+1:

CLR/Libraries/Windows_Devices/windows_devices_stub.cpp
ColinA-MSFT: This is copy-pasted from the SPOT stubs.  I'll change this to CLR_E_NOT_SUPPORTED, though it does mean we'll be inconsistent.

Framework/Core/Windows/Devices/Enumeration/DeviceInformation.cs
ColinA-MSFT: Already done. :)

Framework/Core/Windows/Devices/Enumeration/DeviceInformation.cs
mortezag: I do not either insist whether or not we follow StyleCop, but if it is StyleCop, it has the following rule:  "SA1308 : CSharp.Naming : Variable names must not start with 'm_' or 's_'."

Framework/Core/Windows/Devices/I2c/I2cDevice.cs
ColinA-MSFT: Defined behavior is that an ACK with 0 bytes transferred is distinct from a NACK.  However, this is an extreme edge case, and unlikely to happen in the wild.  The result of the operation is the same (no bytes transferred).  Note that we do check for an intentional 0 transfer before reporting this error (bytesRequested == 0).

Our HAL layer doesn't expose the acknowledgement status cleanly, and the SPOT layer I'm building on here doesn't expose it at all.  I've changed this to FUTURE, as it's an edge case we don't want to spend too much time on now.

smaillet-ms: How can you possibly have an ACK with 0 bytes transferred? AFAIK I2C doesn't have 0 data length transactions.

smaillet-ms:
Can we split this up so we have ALL of GPIO (including the things currently marked TODO:) done, then get I2C, then SPI, etc.. each as separate commits so it's focused on committing one complete thing at a time? There's a lot of different stories here for a single commit.

ColinA-MSFT:
Re-splitting these out for separate changes adds more work than I think it's worth at this point.  Future features will be checked in per-namespace.  The only TODOs left here are individually tracked by issues in GitHub,  Those deltas can come in as bug fixes in individual changes.

PR #100, opened 2015-05-10
Vs2015

smaillet-ms:
- Updated to support Vs2015
- publishing as a new branch since this updates build_sdk to require VS2015 to build the SDK and framework binaries
- added support for STELEM/LDELEM to support code generated by new VS2015 compiler
- disabled VB support
- removed DPWS support as it was broken in a prior release and no-one noticed, indicating a very low usage in the field.
- moved XmlWriter class from DPWS into both the System.XML and System.Xml.Legacy assemblies to allow existing tests to run with minimal modifications
- Fixed and eliminated or suppressed many additional warnings found by the new VS2015 compilers


CLR/Core/CLR_RT_HeapBlock.cpp
```diff
@@ -1897,21 +1897,6 @@ HRESULT CLR_RT_HeapBlock::TransparentProxyValidate() const
...
-{
-    NATIVE_PROFILE_CLR_CORE();
-    DebugBreak();
-}
-
-#else
-
-static void MyAssert() {}
-
-#endif
```
ColinA-MSFT: Nit-pick: In general, we should tag #else/#endif with the original condition.  e.g. "#endif // defined(PLATFORM_WINDOWS) || defined(PLATFORM_WINCE)"

CLR/Core/TypeSystemLookup.cpp
```diff
@@ -1025,13 +1025,12 @@ const CLR_RT_OpcodeLookup c_CLR_RT_OpcodeLookup[] =
     },
     {
         OPDEF(CEE_LDELEM,                     "ldelem",           PopRef+PopI,        Push1,       InlineType,         IObjModel,   1,  0xFF,    0xA3,    NEXT)
-
-        OPDEF_EX(LO_Unsupported, DATATYPE_FIRST_INVALID, 0, CLR_RT_OpcodeLookup::COND_BRANCH_THROW)
+		OPDEF_EX(LO_LoadElement, DATATYPE_FIRST_INVALID, 0, CLR_RT_OpcodeLookup::COND_BRANCH_NEVER | CLR_RT_OpcodeLookup::ATTRIB_HAS_TOKEN)
```
ColinA-MSFT: Mixed tabs and spaces.

DeviceCode/Drivers/LargeBuffer/Test/LargeBuffer_test.cpp
```diff
@@ -15,7 +15,7 @@ void LargeBuffer_PostEvent( UINT16 marshalId )
 
 void LargeBufferTest_Completion(void* arg)
 {
-    LargeBuffer_PostEvent( (UINT16)arg );
+    LargeBuffer_PostEvent( (UINT16)(UINT32)arg );
```
ColinA-MSFT: Should leave a comment why we're doing this.  I assume this gets the compiler to emit different code, but it's very unusual to cast twice.

Framework/Core/System/Xml/XmlWriter.cs
```diff
@@ -802,7 +770,7 @@ public void WriteStartAttribute(string prefix, string localName, string ns)
 
             string localPrefix = "";
 
-            if(prefix != null && prefix.Length == 0) prefix = null;
+            if (prefix != null && prefix.Length == 0) prefix = null;
```
ColinA-MSFT: Suggestion: While you're touching these lines, add braces and move the block to its own line.

Framework/Core/Xml/System/Xml/XmlWriter.cs
```diff
@@ -0,0 +1,1030 @@
...
+        //     The state of writer is not WriteState.Element or writer is closed.
+        //
+        //   System.ArgumentException:
+        //     The xml:space attribute value is invalid.
+        public void WriteAttributeString(string localName, string ns, string value)
+        {
+            this.WriteAttributeString(null, localName, ns, value);
+        }
+
+        //
```
ColinA-MSFT: Can we use a quick regex to format these comments as XML documentation?  This should cause them to carry forward to the assemblies and show up in Intellisense.

ColinA-MSFT:
:+1:

CLR/Core/CLR_RT_HeapBlock.cpp
smaillet-ms: No, we shouldn't be doing that. It just makes for more noise and is often wrong when the logic of the condition changes. But it doesn't matter here as this is code being *DELTETED*

ColinA-MSFT: Ah, brain malfunction, please disregard.  Missed the minuses.  My monitors hide the color differences.

Framework/Core/Xml/System/Xml/XmlWriter.cs
miloush: I did go through most of the code and the comments are often not up to the quality and conventions. I have noticed we had quite nice documentation for [MF 3.0](https://msdn.microsoft.com/en-us/library/hh424067.aspx) - how was this generated? We could use it to generate the documentation back into code.

smaillet-ms: The build systems for the documentation have gone through several iterations, unfortunately when the project moved to our current team the docs system didn't come with it. 
NOTE: that there was nothing new in this code, it's literally a copy of the old file from the previous releases.

miloush: Well for example the [NameTable](https://msdn.microsoft.com/en-us/library/hh422735.aspx) says _Implements a single-threaded XmlNameTable_ while the comment in the file says _XmlNameTable implemented as a simple hash table_ which made me think this is not the code the documentation was generated from.

Or XmlNamespace.NewNamespaceExists current comment suggests a XmlException is thrown if the prefix is alread [sic] in the list (and ArgumentNullExecption if it is null), while the method clearly does not.

So do you want just to transform this into xml documentation as is?

ColinA-MSFT: If it's low-hanging fruit, yes.  It may also be worth auditing the comments for accuracy (as you've done with the examples above), but this is probably quite a bit more expensive to do on a broad scale.

smaillet-ms: FYI this entire pull request was closed and broken down to smaller ones with some of the VS2015 specific stuff completely dropped as we've found we need to take a different path on that. (Additional compiler code generation problems mean even more work to deal with transforming the IL so we're going back to the drawing board - mostly fallback will be to skip the new compiler and leverage the "older" one. When we move to native AOT we'll do it with the new compiler as our front-end from the start.

miloush:
@smaillet-ms  • disabled VB support

Seems that all the compilation errors with Roslyn for VB come from the My template, which makes no sense on device as is anyway (MyProject class + Application, Computer, User and WebServices), unless we have special MF one. This can be disabled however, by defining `_MYTYPE="Empty"`, and to get rid of the *could not find library Microsoft.VisualBasic.dll* BC2017 error, `/vbruntime-` compiler switch works. Since we are referencing our Microsoft.VisualBasic.dll explicitly, it shouldn't be a problem.

smaillet-ms:
@miloush This entire pull request was closed and split into multiple pull requests with additional changes.

miloush:
@smaillet-ms This was the only occurrence talking about disabling VB I found.

smaillet-ms:
@miloush Yes, part of the split was to take a different approach to the problems with VB, and more particularly, Roslyn compilers in general. The solution we are currently running with is to bypass the Roslyn compilers completely for both C# and VB, and using the previous compiler from the v4.x framework. Either way this entire request was closed and never merged, meaning it was abandoned.

miloush:
@smaillet-ms Yes, noticed with #116. Anyway, I guess we would need (and want) to switch to Roslyn eventually at some point, so I thought mentioning what made it compile for me could be helpful - and this was the only relevant place to share it, sorry!

rockybooth:
@miloush, @smaillet-ms It's not clear to me if there is a solution to running NETMF VB 4.3 on Visual Studio.  I explored this here: https://www.ghielectronics.com/community/forum/topic?id=19993&page=1#msg197452 with @mcalsyn.  Visual Studio 2015 will allow creation of NETMF/VB code, so might one consider some kind of fix so others don't waste time?

martincalsyn:
fwiw, I have a working solution based just on vbproj changes (much the same as what miloush found) that allows building VB-based NETMF solutions in 2015. There are a couple changes to the template and then adding STAThreadAttribute (preferably to Microsoft.VisualBasic.dll). Do you want a new pull req with these changes?

smaillet-ms:
We still force the older compiler for both VB and c# as the Roslyn based compilers will generate code assuming generics support exists.

martincalsyn:
Ok - problem solved either way - thanks

smaillet-ms:
What problem is it that is solved? We have support for building in VS2015 - what bug is it that you are talking about fixing here?

martincalsyn:
The problem is with the VB templates not working under VS2015 and the 4.3 QFE2 SDK. I have a workaround (a series of vbproj changes and replacement attribute class) that fix the matter, and I have published that workaround at GHI. What I was discussing here is whether there was any need for those fixes here in 4.4, but since you've fixed that by going back to the old compiler, then we're all good for 4.4. People still need the workaround for 4.3 under VS2015, but unless you are planning to do anything further with 4.3, the workaround will have to suffice and there's nothing more to do.

PR #103, opened 2015-05-12
Removing DPWS

smaillet-ms:
- Removed DPWS support
- removed unused (e.g. dead) code from CLR_RT_HeapBlock
- Fixed VS2015 VSIX generation when building SDK
- added XmlWriter class to System.XML and System.Xml.Legacy to enable existing tests and code using it to run with minimal changes. (Can't just pick one or the other as some tests are for one, and others for the other and the XmlWriter used to be in one of the DPWS assemblies.)
- Disabled FastCompile by default as it screws up the debug info by setting the source line info to that of the preprocessed fastcompile.cpp and not the oringal source location. Furthermore, it wasn't needed anymore with the newer licensing schemes in place for the ARMCC compiler.
- added git ignore entry to ignore temporary directories created by uVision debugger which are essentially a GUID.


mortezag:
:+1: 
(It is really difficult to review one large commit which includes a collection of different changes. I believe it is a better practice to send separate pull requests or at least put them in separate commits)

smaillet-ms:
Unfortunately this is the reduced version of the VS2015 request earlier. It can't really be separated, It's a Mostly a large amount of deletes that are cross dependent. (e.g. pulling only one will break the others)

PR #107, opened 2015-05-13
Add Fault exception clause handling to Metadataprocessor

### LINKED ISSUE #105: Using yield in foreach generates CLR_E_FAIL from MetadataProcessor
From CodePlex issue [2095](http://netmf.codeplex.com/workitem/2095)

When **yield return** statement is in a **foreach** block compiler fails with
 MMP : error MMP0000: CLR_E_FAIL.

 Here is a sample. Tried on both 4.2 and 4.3 sdk versions.
```C#
public static IEnumerable Fails()
{
    var array = new object[] { 1, 2, 3 };
     foreach (var a in array)
         yield return a;
}
```

smaillet-ms:
Added COR_ILEXCEPTION_CLAUSE_FAULT support to assembly linker

Fixes #105


mortezag:
:+1:

PR #110, opened 2015-05-14
Fix whitespace issues in CLR and TinyBooter.

ColinA-MSFT:
- Convert tabs to spaces in files with mixed tabs/spaces.
- Convert mixed line endings to Windows-style (\r\n).

This is not by any means comprehensive for the code base.  These are just the instances which were cheap to find and fix.

### DIFF
--- a/CLR/Libraries/SPOT_Graphics/spot_graphics_native_Microsoft_SPOT_Bitmap.cpp
+++ b/CLR/Libraries/SPOT_Graphics/spot_graphics_native_Microsoft_SPOT_Bitmap.cpp
@@ -733,11 +733,12 @@ HRESULT Library_spot_graphics_native_Microsoft_SPOT_Bitmap::TileImage___VOID__I4
 
     if (bitmap != bitmapSrc)
     {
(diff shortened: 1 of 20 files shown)


CLR/Libraries/SPOT_Graphics/spot_graphics_native_Microsoft_SPOT_Bitmap.cpp
```diff
@@ -766,7 +767,7 @@ HRESULT Library_spot_graphics_native_Microsoft_SPOT_Bitmap::TileImage___VOID__I4
                 dst.right  = dst.left + w - 1;
                 dst.bottom = dst.top + h - 1;
 
-     	        bitmap->DrawImage( dst, *bitmapSrc, src, opacity );
+                bitmap->DrawImage( dst, *bitmapSrc, src, opacity );
```
ColinA-MSFT: No changes here outside line endings.  The diff appears to be confused by the mixed line endings being fixed.  Same for below.

mortezag:
:+1:

PR #112, opened 2015-05-18
fix System.XML assembly checksum

mortezag:
It seems that the current checksum is incorrect, so this change aims to fix the checksum.


smaillet-ms:
Do all the XML tests still pass after fixing this? Usually there are additional changes required as the checksum is based on a "mangled" name of ALL methods of ALL types in the assembly so if the checksum required changing it's likely the actual interop method tables need to be as well. This particular assembly was one involved in the weird corrupted local repo we thought we had taken care of so we should verify those tests before committing this one.

PR #113, opened 2015-05-19
Update Framework and SDK build to VS2015 RC

smaillet-ms:
This update shifts the framework and SDK build to require VS2015 RC (theoretically RTM as well but that will need to be tested once available). After this commit, previous versions of VS will no longer be supported. 

Updates:
- Updated Build scripts to use VS2015 as default compiler for building the framework and SDK
- Redirected VB and C# compilers to use the pre-Roslyn compiler as Roslyn generates code assuming a full V4.0 runtime (including Generics) which makes compiling for NETMF difficult.
- silenced multiple bening warnings detected by newer VS2015 C++ compiler.
- silenced a number of warnings inC# code and tests


mortezag:
:+1:

mortezag:
:+1:

PR #116, opened 2015-05-20
Vs2015 sdk build

smaillet-ms:
- removed support for older VS versions (SDK supports VS2013 and VS2014), building the SDK requires at least VS2015
- updated SDK targets files to allow building applications from VS2015 (disabling Roslyn compiler and falling back to v4.0 compiler as Roslyn generates too much incompatible code to manage easily with re-writers)


mortezag:
:+1:

PR #117, opened 2015-05-21
Build SDK using only VS14 

smaillet-ms:
- Fixed comment type in BuildHelper\dotnetmf.proj
- Upgraded VS integration projects to use VS14
- Fixed setenv_base.cmd to allow building VS2013 VSIX package


mortezag:
:+1:

PR #118, opened 2015-05-21
removed MCBSTM32E solution as it isn't currently maintained or supported

smaillet-ms:
removed MCBSTM32E solution as it isn't currently maintained or supported


mortezag:
:+1:

PR #121, opened 2015-05-21
Removing legacy build support to eliminate confusion

smaillet-ms:
- removed RVDS targets and settings as MDK and RVDS now share the same compiler
- removed unsupported setenv_xxx scripts
- removed unsupported targets from setenv_base
- updated setenv_base so that setenv_mdk will work with default install location of current MDK tools (e.g. MDK_v5 or MDK) non-standard installations still require the path as a paramter to the setenv_mdk script.


mortezag:
:+1:

PR #122, opened 2015-05-22
Remove SolutionWizard

mortezag:


cw2:
Could you please explain why is SolutionWizard being removed? Are there ongoing changes in the NETMF codebase that render it obsolete? If yes, could you please provide more information on what is happening?

Not that SolutionWizard is perfect, but still better than copy'n'paste followed  by 'replace in files' and tedious hand-editing to fix the remaining problems (guid-s, paths etc.)

smaillet-ms:
It's broken in many places. Supporting the requirements of the LWIP stack with an OS along with all the particulars that requires means a significant amount of gutting and hard coding special cases that the whole thing becomes un-maintainable. In it's current form it's just not workable any more.

cw2:
Oh, I see. Thanks for the explanation.

smaillet-ms:
We could leave it around, but it wouldn't "just" work and that would likely cause people problems. Those of us who have been around NETMF a while would be able to tolerate that, but somebody new would get pretty frustrated/annoyed fast. The problem it is trying to salve isn't an easy one, over the years many solutions have been tried across various platform building tools from MS and other companies. This is an area where the industry, as a whole continues to evolve. (For example CMSIS includes the CMSIS-PACK and pack description files to manage dependencies better than what solution wizard and related project tags did - rewriting everything to that model would be a huge amount of work for a minor ".x" release but it's worth considering for the next major release)

cw2:
Ok, I understand (and agree). Should the original comment include the reasons (e.g. *"...because it cannot handle LWIP/RTOS/CMSIS/whatever and/or we cannot support it"*), I would not freak out that much. Coincidentally, I am working on a solution, with a few more in the queue :relieved:

Thanks again for the clarification.

smaillet-ms:
Well, fortunately we are keeping the underlying object model used by SolutionWizard and of course, in version control... nothing is ever really deleted :smile:

mortezag:
@smaillet-ms :+1:?

smaillet-ms:
:+1:

Issue #127, opened 2015-05-22
VSIX package for VS12 (2013) is not built (?)


cw2:
VSIX package for VS 2013 is not built (warning NETMF: Visual Studio 2013 SDK not found VSIX package will not be generated), because `VSSDK12INSTALLDIR` variable is empty - its initialization was removed from `setenv_sdk` in commit 4643ab8a1129006546235eecdf803d7ac62fde5a. Is this intentional?

smaillet-ms:
Not intentional but fixed in a subsequent commit

cw2:
Sorry, but it was broken again in subsequent commit - you can see in the current version the lines are missing.

smaillet-ms:
Fixed in #130

PR #130, opened 2015-05-22
Fixed SDK building VS2013 VSIX #127

smaillet-ms:
fixed set_envbase.cmd to set the location of the VSSDK tools for issue #127


mortezag:
:+1:

PR #135, opened 2015-05-29
Add SPI support (Windows.Devices.Spi) to the Windows.Devices assembly.

ColinA-MSFT:
There's some work left to do here around modifying the HAL layer to report min/max frequencies (currently stubbed as "not implemented"), but the bulk is working and similar in function to Microsoft.SPOT.Hardware.SPI.


miloush:
How is the SPI chip select line working with our current ability to specify any GPIO as a chip select?

CLR/Libraries/Windows_Devices/windows_devices_native_Windows_Devices_Spi_SpiBusInfo.cpp
```diff
@@ -0,0 +1,16 @@
...
+
+#include "windows_devices.h"
+
+HRESULT Library_windows_devices_native_Windows_Devices_Spi_SpiBusInfo::get_ChipSelectLineCount___I4( CLR_RT_StackFrame& stack )
+{
+    TINYCLR_HEADER();
+
+    CLR_RT_HeapBlock* pThis = stack.This(); FAULT_ON_NULL(pThis);
+
+    SetResult_INT32( stack, static_cast<INT32>(CPU_SPI_PortsCount()) );
```
lt72: any reason why not:
stack.SetResult_I4( static_cast<INT32>(CPU_SPI_PortsCount())  ); 
?
generally speaking, the interop API was written for external customers that do not understand the CLR_RT_StackFrame object model.
a piece of small value when we were dealing with sub-par partners...

ColinA-MSFT: No reason, just inconsistent patterns copied within our own code.  Moved to the suggested pattern.

ColinA-MSFT:
SPI chip-select works identically to the existing SPOT API, though the Windows.Devices API makes this a bit less intuitive.

CLR/Libraries/Windows_Devices/windows_devices_native_Windows_Devices_Spi_SpiDevice.cpp
```diff
@@ -0,0 +1,239 @@
...
+        TINYCLR_SET_AND_LEAVE(CLR_E_OBJECT_DISPOSED);
+    }
+
+    settings = pThis[ FIELD__m_settings ].Dereference(); FAULT_ON_NULL(settings);
+    chipSelectPin = settings[ SpiConnectionSettings::FIELD__m_chipSelectionLine ].NumericByRef().u4;
+
+    // Get the pin IDs for this device's chip select line. Then reserve all pins which have assigned IDs.
+    CPU_SPI_GetPins( pThis[ FIELD__m_spiBus ].NumericByRef().u4, mskPin, misoPin, mosiPin );
+
+    if (chipSelectPin != GPIO_PIN_NONE)
```
lt72: if(

CLR/Libraries/Windows_Devices/windows_devices_native_Windows_Devices_Spi_SpiDevice.cpp
```diff
@@ -0,0 +1,239 @@
...
+        SpiMode_Mode1,
+        SpiMode_Mode2,
+        SpiMode_Mode3,
+};
+
+HRESULT Library_windows_devices_native_Windows_Devices_Spi_SpiDevice::InitNative___VOID( CLR_RT_StackFrame& stack )
+{
+    TINYCLR_HEADER(); hr = S_OK;
+
+    GPIO_PIN chipSelectPin = GPIO_PIN_NONE;
```
lt72: typically declarations and operations are aligned on the operator sign: e.g. (not sure that the markup will show this...)

 CLR\Libraries\... cpp files

same applies throughout the file

CLR/Libraries/Windows_Devices/windows_devices_native_Windows_Devices_Spi_SpiDevice.cpp
```diff
@@ -0,0 +1,239 @@
...
+        case SpiMode_Mode3:
+            config.MSK_IDLE = TRUE;
+            config.MSK_SampleEdge = TRUE;
+            break;
+        }
+
+        int dataBitLength = settings[ SpiConnectionSettings::FIELD__m_dataBitLength ].NumericByRef().s4;
+        if (dataBitLength > 8)
+        {
+            config.MD_16bits = TRUE;
```
lt72: mmmh, MD_16bits = TRUE, but calling CPU_SPI_nWrite8_nRead8
that does not sound right, or did something changes in the driver model?

CLR/Libraries/Windows_Devices/windows_devices_stub.cpp
```diff
@@ -59,3 +59,31 @@ HRESULT Library_windows_devices_native_Windows_Devices_Gpio_GpioPin::SetDriveMod
     TINYCLR_SET_AND_LEAVE(CLR_E_NOT_SUPPORTED);
     TINYCLR_NOCLEANUP();
 }
+
+HRESULT Library_windows_devices_native_Windows_Devices_Spi_SpiBusInfo::get_ChipSelectLineCount___I4( CLR_RT_StackFrame& stack )
+{
+    TINYCLR_HEADER();
+    TINYCLR_SET_AND_LEAVE(CLR_E_NOT_SUPPORTED);
+    TINYCLR_NOCLEANUP();
```
lt72: perfectly fine, consider implementing a helper method in the CLR_RT_StackFrame along the lines of:



HRESULT CLR_RT_StackFrame::NotImplementedStub()
{
    NATIVE_PROFILE_CLR_CORE();
    TINYCLR_HEADER();

    if(m_call.m_target->retVal != DATATYPE_VOID)
    {
        SetResult_I4( 0 );
    }

    TINYCLR_SET_AND_LEAVE(CLR_E_NOTIMPL);

    TINYCLR_NOCLEANUP();
}

it leaves the managed stack and managed object cleaner in which it sets the result to a nice round 0

Framework/Core/Windows/Devices/Spi/SpiDevice.cs
```diff
@@ -0,0 +1,285 @@
...
+
+        /// <summary>
+        /// Gets the unique ID associated with the device.
+        /// </summary>
+        /// <value>The ID.</value>
+        public string DeviceId
+        {
+            get
+            {
+                lock (m_syncLock)
```
lt72: because m_deviceId is readonly and m_disposed can only be set once to true, then the lock can be removed
it could be removed anyway because setting a string is a native operation and therefore naturally synchronized on a single threaded system

Framework/Core/Windows/Devices/Spi/SpiDevice.cs
```diff
@@ -0,0 +1,285 @@
...
+
+        /// <summary>
+        /// Gets the connection settings for the device.
+        /// </summary>
+        /// <value>The connection settings.</value>
+        public SpiConnectionSettings ConnectionSettings
+        {
+            get
+            {
+                lock (m_syncLock)
```
lt72: lock can be removed here as well, because if another thread is disposing this object, creating the settigns or not will result in a ObjectDisposedException anyway when the settings are used...
the only thing you are loosing is the causaility order (what happened first) which does not look that interesting...

Framework/Core/Windows/Devices/Spi/SpiDevice.cs
```diff
@@ -0,0 +1,285 @@
...
+
+        /// <summary>
+        /// Releases internal resources held by the device.
+        /// </summary>
+        /// <param name="disposing">True if called from Dispose, false if called from the finalizer.</param>
+        private void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                lock (m_syncLock)
```
lt72: you can remove the lock
the only called of this one is already locking (Dispose()) and the other caller is on a system thread that knows better

Framework/Core/Windows/Devices/I2c/I2cDevice.cs
```diff
@@ -32,6 +32,7 @@ public sealed class I2cDevice : IDisposable
         private static object s_deviceLock = new object();
```
lt72: RE: thread safety on I2C device. I noticed that there is static lock to implement thread safety for the object. Is that a UWP requirement? If not, consider removing, because no other HW type in NETMF is thread safe anyway.

CLR/Libraries/Windows_Devices/windows_devices_native_Windows_Devices_Spi_SpiDevice.cpp
ColinA-MSFT: We're pretty inconsistent about this one, and I kept having to fight VS to do it.  I think we should stick with the more traditional formatting unless you feel very strongly.

CLR/Libraries/Windows_Devices/windows_devices_stub.cpp
ColinA-MSFT: I was using this earlier, but Steve asked me to switch to not supported... I'll switch it back.

Framework/Core/Windows/Devices/Spi/SpiDevice.cs
ColinA-MSFT: Will remove all synchronization.  The UWP version is free-threaded, I believe, which generally mandates that all operations be thread safe.  However, since this can't match 100% anyway, it's likely not worth the overhead.

CLR/Libraries/Windows_Devices/windows_devices_native_Windows_Devices_Spi_SpiDevice.cpp
ColinA-MSFT: We have to fight the Visual Studio defaults on this one, and the code base is inconsistent.  Pushing back unless you feel very strongly about it.

smaillet-ms:
:+1:

PR #137, opened 2015-06-02
Fixed build break of not enough LR_Dat memory spaces and updated the comments

ppjcheng:
Increase the size of LR_dat and updated comments.


smaillet-ms:
:+1:

PR #139, opened 2015-06-03
Fix DateTime.Ticks to match the behavior of .NET

### LINKED ISSUE #138: DateTime - Difference in Epoch between .NET and .NET Micro
Refer to http://netmf.codeplex.com/workitem/2422

In .NET, DateTime.Ticks represents the number of 100-nanosecond intervals that have elapsed since 12:00:00 midnight, January 1, 0001. However, in .NET Micro, DateTime.Ticks returns the number of ticks since 12:00 AM January 1, year 1601 A.D.

mortezag:
This pull request fixes #138 

Basically, the interface of DateTime type is modified so that Ticks correspond as the number of 100-nanosecond intervals that have elapsed since 12:00:00 midnight, January 1, 0001 instead of 12:00 AM January 1, year 1601 A.D.
This is done without modifying the native portion of DateTime to minimize the required changes.

### DIFF
--- a/Framework/Subset_of_CorLib/System/DateTime.cs
+++ b/Framework/Subset_of_CorLib/System/DateTime.cs
@@ -35,7 +35,17 @@ namespace System
      * This value type represents a date and time.  Every DateTime
      * object has a private field (Ticks) of type Int64 that stores the
      * date and time as the number of 100 nanosecond intervals since
-     * 12:00 AM January 1, year 1601 A.D. in the proleptic Gregorian Calendar.
+     * 12:00 AM January 1, year 1601 A.D. in the proleptic Gregorian Calendar. 
+     * See DeviceCode\PAL\time_decl.h for explanation of why we are taking
+     * year 1601 as origin for our HAL, PAL, and CLR.
+     *
+     * It should be considered that previously DateTime.Ticks used to represent
+     * the number 100 nanosecond intervals since 12:00 AM January 1, year 1601 A.D.
+     * causing a 504911232000000000 difference between ticks in .NET and .NET MF
+     * To fix this disparity, we modified the CLR portion of DateTime type without 
+     * changing the origin of the DateTime. In particular, TicksAtOrigin is subtracted
+     * from any ticks value provided by user code, and it is added back whenever the 
+     * ticks value is read.
      *
      * <p>For a description of various calendar issues, look at
      * <a href="http://serendipity.nofadz.com/hermetic/cal_stud.htm">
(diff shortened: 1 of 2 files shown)


Framework/Subset_of_CorLib/System/DateTime.cs
```diff
@@ -93,13 +88,18 @@ public struct DateTime
         private const ulong TickMask = 0x7FFFFFFFFFFFFFFFL;
         private const ulong UTCMask = 0x8000000000000000L;
 
-        public static readonly DateTime MinValue = new DateTime(MinTicks);
-        public static readonly DateTime MaxValue = new DateTime(MaxTicks);
+        // Ticks at 12:00 AM January 1, year 1601 A.D.
```
smaillet-ms: It's worth putting a more detailed comment here explain why this is done this way. (e.g. shifting to compat with desktop without breaking existing HAL). We'll also need to ensure we highlight this behavioral change in our release notes for this version.

Test/Platform/Tests/CLR/mscorlib/systemlib/systemlib2/SystemDateTimeTests.cs
```diff
@@ -1941,11 +1944,11 @@ public MFTestResults DateTime_TicksTest52()
             try
             {
                 Log.Comment("Creating a DateTime, getting the Ticks and Verifying");
-                DateTime testDateTime = new System.DateTime(0);
+                DateTime testDateTime = new System.DateTime(635688622030533832);
```
smaillet-ms: Umm, what's with the magic number? Can we get a named const for this?

mortezag: Actually, this is not a magic number but is a random number. The goal of the test as I understood is to make sure when you create a DateTime with a tick value, you get the same value from the Ticks property.

smaillet-ms: OK, so a named const with a comment explaining that would help clarify that point to so no-one else needs to wonder...

smaillet-ms:
:+1:

Framework/Subset_of_CorLib/System/DateTime.cs
miloush: So what is the rationale behind carrying this workaround in managed code and not fixing it in native? What code could potentially break in existing HAL?

mortezag: The native part was designed based on 12:00 AM January 1, year 1601 A.D. as the epoch. Technically, it was also possible to fix it from the native layer but required more changes with a greater amount of regression risk. So we decided to fix it from the managed layer and leave the native part untouched (at least for this version).

miloush: While definitely easier the amount of risk potentially involved wasn't really expressed, which could be helpful to contributors facing the same question whether to fix in managed or native code. Can we perhaps put some guidance on the wiki?

lt72:
We will revert this fix, add guidance in a wiki, and re-open the issue for 4.5.
the correct fix has toe be carried out in Time_decl.h and it is too risky at this point. 
Please also add information on https://github.com/NETMF/netmf-interpreter/issues/264 and close it.

PR #140, opened 2015-06-03
Adding MCBSTM32F400_NONET solution

smaillet-ms:
- Added solution for MCBSTM32F400 that does not use networking or the CMSIS_RTOS unerneath.
- Since the non network uses a different startup and init, refactored the Cortex-Mx init and startup from MCBSTM32F400 and the Targets\STM32F$\CortexM3 locations out to generalized forms
- removed old legacy arm9 CPU support files as we aren't testing or validating them anymore.(if anyone is still building for those cores, they can, of course continue using their existing files)


smaillet-ms:
Closing this request as creating another solution with cloned code is just a bad idea and a maintenance problem. While it's how it was always done in the past current experience on this has shown it makes life harder. The build infrastructure can handle multiple projects in a solution and there is nothing special about the TinyCLR project name (beyond convention) and so we can in fact have a TinyCLR_NONET folder and project inside the same solution so that they can share all of the code between them, no need to copy and keep in synch etc... So I'll be shifting the No networking configuration to use that pattern instead of as a cloned solution.

PR #144, opened 2015-06-03
Delete old lwip stack

mortezag:
As we have a newer/cleaner version of LWIP stack in the tree, we want to remove the old lwip stack.
If you believe the old stack should not be deleted, please let us know.


smaillet-ms:
:+1:

smaillet-ms:
Can we add (or perhaps more appropriately remove) the STM32F4_ETH_LWIP target lib as well since it is effectively orphaned by this now.

ColinA-MSFT:
:+1:

PR #146, opened 2015-06-04
Fix NullReferenceException in String.Concat(params object[] args)

### LINKED ISSUE #145: NullReferenceException in String.Concat(params object[] args)
refer to http://netmf.codeplex.com/workitem/2340

mortezag:
fixes #145


smaillet-ms:
:+1:

PR #149, opened 2015-06-05
Fix Byte.ToString(X2)

### LINKED ISSUE #133: Byte X2 formatting returns 8 bytes for values above 127 in emulator
```csharp
byte b = 255;
string s = b.ToString("X2");
```

This code returns `"FF"` on device but `"FFFFFFFF"` in the emulator (similarly for all byte values above 127). `"FF"` is expected.

mortezag:
Fixes #133

### DIFF
--- a/Framework/Subset_of_CorLib/System/Number.cs
+++ b/Framework/Subset_of_CorLib/System/Number.cs
@@ -371,7 +371,7 @@ namespace System
                     {
                         int len = result.Length;
 
-                        if(value is sbyte)
+                        if(value is sbyte || value is byte)
                         {
                             if (len > 2)
                             {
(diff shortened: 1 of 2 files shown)


Framework/Subset_of_CorLib/System/Number.cs
```diff
@@ -371,7 +371,7 @@ private static String PostProcessInteger(Object value, String original, char for
                     {
                         int len = result.Length;
 
-                        if(value is sbyte)
+                        if(value is sbyte || value is byte)
```
smaillet-ms: Don't we have a similar signed/unsigned problem below for short vs. ushort?

mortezag: It seems it does not.

smaillet-ms:
:+1:

Framework/Subset_of_CorLib/System/Number.cs
miloush: I think this is because this "post processing" is actually fixing wrong native output in the case of byte. Anyone reading this fixed code would question the ushort case too.

Might be worth fixing the original cause, e.g. in [corlib_native_System_Number.cpp](../blob/dev/CLR/Libraries/CorLib/corlib_native_System_Number.cpp) changing `char` to `unsigned char` unless there is a compiler reason not to (works for me in the emulator).
```cpp
case DATATYPE_U1:  
hal_snprintf( result, ARRAYSIZE(result), "%.*X", precision,
                                               (unsigned char)value->NumericByRef().u1 );  
break; 
```

Issue #147, opened 2015-06-04
new StringBuilder().ToString() returns null but it should return ""


mortezag:
new StringBuilder().ToString() on NETMF return null while the same code on .NET return String.Empty. For compatibility between NETMF and .NET behavior, I believe StringBuilder in NETMF should be modified.

PR #150, opened 2015-06-05
Fix new String(new Char[0])

### LINKED ISSUE #148: Difference between String(char[] value) constructor in NETMF and .NET
Instantiating a String object with zero-size char array in NETMF results in a null string while on .NET it results in String.Empty. For compatibility between NETMF and .NET behavior, I believe this constructor in NETMF should be modified.

mortezag:
fixes #148 and #147


smaillet-ms:
:+1:

Issue #26, opened 2015-03-19
Support for recent GCC toolchain and STM32F4-Discovery board


cw2:
I would like to contribute support for recent GCC toolchain ([GNU Tools for ARM Embedded Processors](https://launchpad.net/gcc-arm-embedded), including newlib-nano), but I don't have any MCBSTM32 board for testing, I was thinking something more affordable like STM32F4-Discovery.

Are there any plans to include STM32F4-Discovery port in the official NETMF?

piwi1263:
*I'll vote for this as well, mostly for the support of STM32F4xx MCU with
the most recent GCC toolchain. *

CW2, what MCU is on that board the STM32F427 or the STM32F429 ? There are a
view discovery boards. The nine series is with an LCD controller on board.
Maybe the STM32F411 is a good power efficent MCU alternative. It would be
on an even cheaper nucleo board.



2015-03-19 16:17 GMT+01:00 cw2 <notifications@github.com>:

smaillet-ms:
We don't have any plans for BSPs beyond the MCNSTM32F400, however since it is effectively a superset of a large family of STM32 chips it can be used as the basis for a large set of boards, We are perfectly happy to take additional ports (Including additional chipsets) as contributions

On that note:
One thing we are planning to get going fairly soon. Is better support for CMSIS and mBed HAL libraries that already exist. This would allow re-use of the existing "driver" libraries for a broader range of chips without the need to re-write or wrap them just for NETMF. This would make it easier to port to the majority of different Cortex-M chips,

cw2:
@piwi1263 Initially, I am going to use [STM32F4-Discovery](http://www.st.com/web/catalog/tools/FM116/SC959/SS1532/LN1848/PF252419) that has STM32F407VGT6. Then, it depends - if there is enough interest and resources... but my personal focus would be more likely towards smaller micros.

cw2:
@smaillet-ms Thanks for the info. Regarding the driver libraries, that is something I have been exploring too...

ktos:
NETMF for Discovery4 is available now and is working - I'm using this board with STM32F407VG and NETMF 4.3.1 using patches from https://netmf4stm32.codeplex.com/ (Discovery4 itself) and https://stm32f4netmfgcc.codeplex.com/ (GCC patches) compiled with GCC 4.6.2. And a bit of small patches by myself if I remember, but it works - except Tinybooter, I cannot compile Tinybooter with GCC (to work).

[...]

PR #154, opened 2015-06-07
Added barebone STM32F4DISCOVERY solution

cw2:
This is a minimal STM32F4DISCOVERY solution, for #26. It is based on MCBSTM32F400_NONET, so no RTOS.

The MDK build seems to be working fine, I have verified TinyBooter commands in MFDeploy, flashing TinyCLR, application deployment and debugging from Visual Studio.

The GCC build (using the latest GCC ARM Embedded Tools 4.9.3) has working TinyBooter, but there is a problem with TinyCLR, which can be flashed fine and runs, but there seems to be some memory corruption or invalid settings, because the MFDeploy "Memory Map" command returns wrong result for RAM  (the reported address is in flash range). I suspect I overlooked something...

Notes:
* The bootloader mode is indicated by turning on all four user LEDs
* The User button does not work yet, so it is not possible to switch to bootloader on startup (sorry).
* GCC uses newlib-nano, generated code is about the same size as produced by MDK, i.e. the flash memory layout stays the same (scatterfile, BlockRange).
* Code from pull request #153 is already included.


Stabilitronas:
Wow, this is so cool! Good job!

smaillet-ms:
I'd recommend comparing and updating this code base to the latest MCBSTM32F400 support in this branch, a number of issues were resolved there that don't appear in this new cloned solution.

cw2:
Could you please be more specific on what is not included? The branch is based on commit 7efe0092892f1b4ac977cdfd2dc6901b898d9040 from 6/6/2015 6:31:47 PM and there are no additional changes in MCBSTM32F400 solution after that (?) Or is the pull request wrong?

smaillet-ms:
you based this on MCBSTM32F400_NONET, which no longer exists. A number of changes were made to the GCC support and the Keil board solution. (And there's still more to go) so it's probably worth double checking to see if any of the changes matter.

cw2:
I have fixed the RAM layout and STM32F4DISCOVERY solution build by GCC seems to be working: flashing TinyCLR from bootloader, MFDeploy commands, application deployment and debugging from Visual Studio.

smaillet-ms:
:+1:

PR #159, opened 2015-06-09
Fix Single.ToString() for special values

### LINKED ISSUE #157: Single.ToString not consistent for Infinity and NaN
refer to http://netmf.codeplex.com/workitem/1848

mortezag:
Fix ToString method for float type so that it returns correct strings for NaN, Infinity, and -Infinity values

Fixes #157

### DIFF
--- a/Framework/Subset_of_CorLib/System/Single.cs
+++ b/Framework/Subset_of_CorLib/System/Single.cs
@@ -22,14 +22,34 @@ namespace System
 
         public override String ToString()
         {
-            return Number.Format(m_value, false, "G", NumberFormatInfo.CurrentInfo);
+            // Number.Format method is responsible for returning the correct string representation of the value; however, it does not work properly for special values.
+            // Fixing the issue in Number.Format requires a significant amount of modification in both native and managed code.
+            // In order to avoid that (at lease for now), we use the help of Double class to identify special values and use Number.Format for the others.
+            string str = ((Double)m_value).ToString();
+            switch (str)
+            {
+                case "Infinity":
+                case "-Infinity":
+                case "NaN":
+                    return str;
+                default:
+                    return Number.Format(m_value, false, "G", NumberFormatInfo.CurrentInfo);
+            }
         }
 
         public String ToString(String format)
         {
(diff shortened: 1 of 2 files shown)


Framework/Subset_of_CorLib/System/Single.cs
```diff
@@ -22,14 +22,31 @@ public struct Single
 
         public override String ToString()
         {
-            return Number.Format(m_value, false, "G", NumberFormatInfo.CurrentInfo);
+            string str = ((Double)m_value).ToString();
```
smaillet-ms: This seems kinda wonky.... Shouldn't the fic be done in Number.Format directly instead of here? Isn't it possible something else calls Number.Format() that would still get the incorrect behavior?

mortezag: AFAIK, only Double and Single are using Number.Format method with these parameters. Double class has its own mechanism to find out special values and treat them accordingly. Double also does it at the managed layer (with the help of a few native functions). For the Single type, we do not have those native methods and implementing them is not trivial, so I got help from the Double type.
I agree the it would be better if you could fix it at the native layer but it requires a significant amount to effort (implementing a bunch of native method to recognize special values for the Single type) 
Similarly, bringing the logic to Number class requires modifications to Double types while does not add a lot of value.
I believe it is better to keep the change here for now.

smaillet-ms: OK, then let's at least add a comment to explain that point.

Framework/Subset_of_CorLib/System/Single.cs
```diff
@@ -22,14 +22,34 @@ public struct Single
 
         public override String ToString()
         {
-            return Number.Format(m_value, false, "G", NumberFormatInfo.CurrentInfo);
+            // Number.Format method is responsible for returning the correct string representation of the value; however, it does not work properly for special values.
```
smaillet-ms: Can you reflow the comments here, they are really long and don't fit on the screen in a GitHub review...

smaillet-ms:
:+1:

PR #161, opened 2015-06-10
Add support for Windows.Devices.Adc

ColinA-MSFT:
This feature is currently lacking a reference implementation, but adds support for the UWP ability to hook ADC Providers into the richer API.  Reference implementations for IAdcControllerProvider will be provided at a later date. Tracked by issue #163.


ColinA-MSFT:
Note that this change is missing XML doc comments.  These will be added when the official UWP documentation is published.

smaillet-ms:
:+1:

Issue #162, opened 2015-06-10
Update XML doc comments for Windows.Devices


ColinA-MSFT:
XML doc comments for Windows.Devices are largely based on pre-ship documentation for our UWP counterparts.  We should update these when the doc writers update the source MSDN documentation.

PR #166, opened 2015-06-11
Add support for Windows.Devices.Pwm.

ColinA-MSFT:
As with ADC, no reference implementation yet exists.  This will be provided in a separate change.

Also as with ADC, this feature doesn't yet include full XML documentation.  This will be added when final UWP documentation has been published on MSDN.  Tracked by issue #162.


Framework/Core/Windows/Devices/Pwm/PwmController.cs
```diff
@@ -0,0 +1,86 @@
...
+    public sealed class PwmController
+    {
+        private IPwmControllerProvider m_provider;
+        private double m_minFrequency;
+        private double m_maxFrequency;
+
+        internal PwmController(IPwmControllerProvider provider)
+        {
+            m_provider = provider;
+            m_minFrequency = provider.MinFrequency;
```
ColinA-MSFT: Cached these since we compare with them later.  They should remain static.  We may want to also cache other values that are expected to never change, but I don't want to bloat the object too much.

Framework/Core/Windows/Devices/Pwm/PwmEnums.cs
```diff
@@ -0,0 +1,8 @@
+namespace Windows.Devices.Pwm
+{
+    public enum PwmPulsePolarity
+    {
+        ActiveHigh = 0,
```
ColinA-MSFT: The UWP enum is in this order, though typically "Low" is zero.  In this case, it's appropriate since "Active High" is the default.

Framework/Core/Windows/Devices/Pwm/PwmPin.cs
```diff
@@ -0,0 +1,171 @@
...
+namespace Windows.Devices.Pwm
+{
+    public sealed class PwmPin : IDisposable
+    {
+        private readonly int m_pinNumber;
+        private PwmController m_controller;
+        private IPwmControllerProvider m_provider;
+        private bool m_started = false;
+        private bool m_disposed = false;
+        private double m_dutyCycle = 0;
```
ColinA-MSFT: Confirmed with UWP folks: Default behavior is no signal (always low) until overridden by a caller.

Framework/Core/Windows/Devices/Pwm/PwmPin.cs
```diff
@@ -0,0 +1,171 @@
...
+                GC.SuppressFinalize(this);
+                m_disposed = true;
+            }
+        }
+
+        private void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                m_provider.ReleasePin(m_pinNumber);
```
ColinA-MSFT: We don't disable the pin here.  Provider is expected to do that when the pin is released.  This will be documented before release.

Framework/Core/Windows/Devices/Pwm/PwmController.cs
```diff
@@ -0,0 +1,86 @@
+using System;
+using Windows.Devices.Pwm.Provider;
+
+namespace Windows.Devices.Pwm
+{
+    public sealed class PwmController
+    {
+        private IPwmControllerProvider m_provider;
+        private double m_minFrequency;
+        private double m_maxFrequency;
```
mortezag: why you defined m_manFrequency and m_maxFrequency? similar to the other properties, you could get them from m_provider whenever you want them.

ColinA-MSFT: The main reason is that these shouldn't be allowed to change, even if the provider reports a different value in the future.  However, I can remove these cached values if you feel strongly.

mortezag: As you said, the value of these two properties are cached while the value of other properties are not, e.g. pin count. So basically there is a difference between the behavior of these two properties and the others. If this difference is expected keep them as they are; otherwise, change them to match the others (or vice versa).

ColinA-MSFT: Will cache all values (and I'll go back and do the same for ADC).

smaillet-ms: No, remove the cacheing, since these are not expected to change then they aren't needed very often either so the performance overhead of another call to native code isn't as big an issue as storing the data twice. (RAM s a VERY limited resource on these micro controllers so when it comes to tradeoffs the general rule is favor reduced memory usage unless tere is a measured wuantifiable justification otherwise)

ColinA-MSFT: This isn't a performance optimization. It's insulation for Providers from common API violations (altering a property that's intended to be constant).  I've removed the caching anyway, but this does remove our ability to arbitrate intended behavior.

smaillet-ms:
:+1:

Framework/Core/Windows/Devices/Pwm/PwmController.cs
miloush: Low power modes could lower the maximum frequency, couldn't they?

ColinA-MSFT: Providers should report their greatest range (lowest min and highest max).  This API will guard against users setting anything outside that range.  However, more complex behavior, such as variable sub-ranges must be arbitrated by the Provider.

This is the philosophy adopted by the designers of these APIs.  We're trying to mirror those decisions as closely as possible for the sake of cross-device consistency.

PR #169, opened 2015-06-12
Fix UTF8Encoding.UTF8.GetBytes

mortezag:
This overload of the GetBytes method had some fundamental issues. Specifically, it assumed that every character maps to one byte which is not correct for UTF8. This change fix the problem by involving UnicodeHelper.

### DIFF
--- a/CLR/Libraries/CorLib/corlib_native_System_Text_UTF8Encoding.cpp
+++ b/CLR/Libraries/CorLib/corlib_native_System_Text_UTF8Encoding.cpp
@@ -32,24 +32,39 @@ HRESULT Library_corlib_native_System_Text_UTF8Encoding::GetBytes___I4__STRING__I
     NATIVE_PROFILE_CLR_CORE();
     TINYCLR_HEADER();
 
-    size_t                  cMaxBytes;
     LPCSTR                  str         = stack.Arg1().RecoverString();
     CLR_INT32               strIdx      = stack.Arg2().NumericByRef().s4;
     CLR_INT32               strCnt      = stack.Arg3().NumericByRef().s4;
     CLR_RT_HeapBlock_Array* pArrayBytes = stack.Arg4().DereferenceArray();
     CLR_INT32               byteIdx     = stack.Arg5().NumericByRef().s4;
 
+    const CLR_UINT8*        i;
+    const CLR_UINT8*        j;
+    CLR_RT_UnicodeHelper    uh;
+    int                     strLength;
+
     FAULT_ON_NULL(str);
     FAULT_ON_NULL(pArrayBytes);
 
-    cMaxBytes = hal_strlen_s(str);    
+    uh.SetInputUTF8(str);
+    strLength = uh.CountNumberOfCharacters();
+    if (strLength < 0) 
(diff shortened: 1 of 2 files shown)


CLR/Libraries/CorLib/corlib_native_System_Text_UTF8Encoding.cpp
```diff
@@ -32,24 +32,36 @@ HRESULT Library_corlib_native_System_Text_UTF8Encoding::GetBytes___I4__STRING__I
...
+    if ((strIdx + strCnt) > (CLR_INT32)strLength) TINYCLR_SET_AND_LEAVE(CLR_E_OUT_OF_RANGE);
+
+    uh.ConvertFromUTF8(strIdx, true);
+    i = uh.m_inputUTF8;
+    uh.ConvertFromUTF8(strCnt, true);
+    j = uh.m_inputUTF8;
 
-    if((strIdx  + strCnt) > (CLR_INT32)cMaxBytes                   ) TINYCLR_SET_AND_LEAVE(CLR_E_OUT_OF_RANGE);
-    if((byteIdx + strCnt) > (CLR_INT32)pArrayBytes->m_numOfElements) TINYCLR_SET_AND_LEAVE(CLR_E_OUT_OF_RANGE);
+    if ((byteIdx + j - i) > (CLR_INT32)pArrayBytes->m_numOfElements) TINYCLR_SET_AND_LEAVE(CLR_E_OUT_OF_RANGE);
```
smaillet-ms: Split this line. While there are a lot of cases of this bad style in the CLR source, we should endeavor to fix them as we make changes to the code base.

Test/Platform/Tests/CLR/mscorlib/systemlib/systemlib2/Utf8EncodingTests.cs
```diff
@@ -76,5 +76,24 @@ public MFTestResults Utf8EncodingTests_Test2()
...
+            bool result = true;
+
+            string str = "AB\u010DAB";
+
+            byte[] data = new byte[4];
+            int count = UTF8Encoding.UTF8.GetBytes(str, 1, 3, data, 0);
+
+            result &= count == 4;
+
+            result &= (new string(UTF8Encoding.UTF8.GetChars(data)) == "B\u010DA");
```
ColinA-MSFT: We should generally use logical && for this type of operation, rather than bitwise &.  We should also probably short-circuit when anything fails, so something like the following (which mirrors the chained "if (SUCCEEDED(hr))"  pattern):

if (result)
{
    result = expr;
}

if (result)
{
    ...
}

mortezag: I agree with your point; however, all the other tests are implemented in this manner. When we start implementing new test suits we should do it right, but for now I think it is better to keep it uniform as we do not have any plan to fix the others.

ColinA-MSFT: Agreed. Consistency trumps my feedback here.

ColinA-MSFT:
:+1:

smaillet-ms:
:+1:

PR #171, opened 2015-06-12
Update uvision proj

smaillet-ms:
- reset keil project doc paths to default install location so they work correctly (unfortunately the uVision project format doesn't distinguish between per project and per user persistent data very well)
- removed preset breakpoints in uVision project as that just confuses multiple users of checked in code
- updated STM32 CMSIS pack  reference in uVision project to latest version
- updated debugger settings to allow ITM "printf" capture in TinyBooter


mortezag:
:+1:

PR #178, opened 2015-06-17
make Microsoft.VisualBasic.Constants actually constant

### LINKED ISSUE #170: Microsoft.VisualBasic.Constants.vbCrLf does is not a constant
In NETMF 4.3 The function:
        Function Something(Optional x As String = Microsoft.VisualBasic.Constants.vbCrLf) As String
            Return x
        End Function
causes an error "Constant expression is required" for Microsoft.VisualBasic.Constants.vbcrlf.

iamin at https://www.ghielectronics.com/community/forum/topic?id=19492&page=1#msg192654 points out that In NETMF this is defined:
public sealed class Constants
{
public static string vbCrLf = "\r\n";
// ...
}

and on full NET this is:
public sealed class Constants
{
public const string vbCrLf = "\r\n";
// ...
}

smaillet-ms:
Fixes #170 
converted "shared" members of Constnts class to Const to match Full framework pattern.


ColinA-MSFT:
:+1:

PR #182, opened 2015-06-17
Fix PWM issues.

ColinA-MSFT:
This addresses two issues from CodePlex:
- Extend the PWMChannel enum to 16 values by default.  These can still be overridden by board manufacturers.
- Fix PWM.Dispose to call the internal Dispose(bool) implementation exactly once.


smaillet-ms:
:+1:

PR #185, opened 2015-06-18
Fixed MDK targetsfile for multi-project solutions

smaillet-ms:
Fixed MDK targets to use $(AssemblyName) instead of hard coded to tinyclr


lt72:
Good to go

Issue #180, opened 2015-06-17
Windows.Devices assembly missing from install package


ColinA-MSFT:
Windows.Devices needs to be added to the install package.  It should include the following files from BuildOutput\public\Release\Client:

dll\Windows.Devices.dll -> %InstallDir%\Assemblies\le\
dll\Windows.Devices.pdb -> %InstallDir%\Assemblies\le\
pe\le\Windows.Devices.pdbx -> %InstallDir%\Assemblies\le\
pe\le\Windows.Devices.pe -> %InstallDir%\Assemblies\le\
dll\Windows.Devices.dll -> %InstallDir%\Assemblies\be\
dll\Windows.Devices.pdb -> %InstallDir%\Assemblies\be\
pe\be\Windows.Devices.pdbx -> %InstallDir%\Assemblies\be\
pe\be\Windows.Devices.pe -> %InstallDir%\Assemblies\be\

PR #187, opened 2015-06-18
Updated SDK installer to include Windows.Devices support

smaillet-ms:
- Added Windows.Devices reference assemblies
- Added Windows.Devices device side LE/BE assemblies
- updated general install merge modules to include only the VS14 merge modules as everything should be building against VS14 now.


ColinA-MSFT:
:+1:

ColinA-MSFT:
Resolves #180.

PR #189, opened 2015-06-18
Fixed delays due to debug port buffer fills when no debugger present.

### LINKED ISSUE #188: USB slowdown due to timeouts if no debugger attached.
Moved from CodePlex [WorkItem:2179](https://netmf.codeplex.com/workitem/2179)

smaillet-ms:
Fix #188 
- added new parameter to DebugPort_Write for max retries with default value of 99 to match previous code behavior unless explicitly set to something else.
- modified all debug prints string functions and calls to DebugPort_Write() to use a retriy count of 0 so that if the first attempt to send the string fails due to buffer full, it just returns and moves on with life.
- fixed tabs in MCPSTM32F400 Hardware managed code CPU class (as that was used for testing this)


mortezag:
:+1:

PR #190, opened 2015-06-19
Set error on LINGER and DONTLINGER socket options

mortezag:
As linger and dontlinger options were hacked into previous lwip stack and those hacks do not exist in the new lwip options, calling setsockopt and getsockopt with these values will cause problems. This fix returns proper errors for them.

### DIFF
--- a/DeviceCode/pal/lwip_1_4_1_os/SocketsDriver/lwIP__Sockets.cpp
+++ b/DeviceCode/pal/lwip_1_4_1_os/SocketsDriver/lwIP__Sockets.cpp
@@ -562,13 +562,28 @@ int LWIP_SOCKETS_Driver::SetSockOpt( SOCK_SOCKET socket, int level, int optname,
             nativeOptionName = GetNativeSockOption(optname);            
 
             switch(optname)
-            {
+            {        
+                // LINGER is not implemented in LWIP
+                case SOCK_SOCKO_LINGER:
+                    if (*(int*)optval)
+                    {
+                        errno = SOCK_ENOPROTOOPT;
+                        return SOCK_SOCKET_ERROR;
+                    }
+                    return 0;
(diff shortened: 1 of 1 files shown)


DeviceCode/pal/lwip_1_4_1_os/SocketsDriver/lwIP__Sockets.cpp
```diff
@@ -562,13 +562,18 @@ int LWIP_SOCKETS_Driver::SetSockOpt( SOCK_SOCKET socket, int level, int optname,
             nativeOptionName = GetNativeSockOption(optname);            
 
             switch(optname)
-            {
+            {        
+                // LINGER and DONTLINGER are not implemented in LWIP
+                case SOCK_SOCKO_LINGER:
+                case SOCK_SOCKO_DONTLINGER:
+                    errno = SOCK_ENOPROTOOPT;
+                    return SOCK_SOCKET_ERROR;
```
smaillet-ms: Any chance we could use a more descriptive error code! This smells a bit like E_FAIL... (same for other cases in here)

mortezag: Maybe I didn't get your point but the error code is pretty specific. The error returned to the user is SOCK_ENOPROTOOPT which means "Bad protocol option: An unknown, invalid or unsupported option or level was specified in a getsockopt or setsockopt call." Unfortunately, it is not possible to return a error message to the user.

smaillet-ms:
OK, I guess I misread the code. the errno is what is exposed to the user, I was concerned about the SOCK_SOCKET_ERROR return value (Which is basically a generic E_FAIL) since that's not really the error code that percolates up to the caller it's effectively a success/fail so no worries.

smaillet-ms:
sorry should have include :+1:

PR #192, opened 2015-06-20
Fix array range exception during WriteConfig

martincalsyn:
In WriteConfig at line 878 in MFConfigHelper.cs, a temp buffer is allocated to include the current config plus the addition.  When the length of the current config plus the new data is larger than the current size of the m_cfg_all_data array, the temp array should be enlarged to accomodate the new data. In the existing code, the allocation uses m_lastCfgIndex instead of _lastCfgIndex+data.Length and so under-allocates space when m_lastCfgIndex+data.Length would exceed the current size of m_all_cfg_data.Length. This fix corrects the under-allocation and resulting exception on line 881.


smaillet-ms:
:+1:

PR #198, opened 2015-06-23
Fix Peek method in StreamReader

### LINKED ISSUE #197: IndexOutOfRangeException ReadLine
refer to http://netmf.codeplex.com/workitem/2048

mortezag:
Fixes #197

### DIFF
--- a/Framework/Core/System/IO/StreamReader.cs
+++ b/Framework/Core/System/IO/StreamReader.cs
@@ -87,7 +87,8 @@ namespace System.IO
             int nextChar;
 
             // If buffer need refresh take into account max UTF8 bytes if the next character is UTF8 encoded
-            if ((m_curBufPos                       == (m_curBufLen-1)) || 
+            // Note: In some occasions, m_curBufPos may go beyond m_curBufLen-1 (for example, when trying to peek after reading the last character of the buffer), so we need to refresh the buffer in these cases too
+            if ((m_curBufPos                       >= (m_curBufLen-1)) || 
                ((m_buffer[m_curBufPos + 1] & 0x80) !=  0 && 
                 (m_curBufPos + 3                   >=  m_curBufLen)))
             {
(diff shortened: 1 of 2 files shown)


Framework/Core/System/IO/StreamReader.cs
```diff
@@ -87,7 +87,7 @@ public override int Peek()
             int nextChar;
 
             // If buffer need refresh take into account max UTF8 bytes if the next character is UTF8 encoded
-            if ((m_curBufPos                       == (m_curBufLen-1)) || 
+            if ((m_curBufPos                       >= (m_curBufLen-1)) || 
                ((m_buffer[m_curBufPos + 1] & 0x80) !=  0 && 
                 (m_curBufPos + 3                   >=  m_curBufLen)))
```
smaillet-ms: Is this another hidden bug lurking? (e.g. >= m_curBufLen instead of m_curfLen-1 like the other case? ) It's certainly confusing to see two forms of edge test cases like that. Especially when the thing compared against in the second case is an addition to a const...

mortezag: The problem is that the read method increment the position pointer on every read. On the next read, if the position is out of range, it reload the internal buffer. This works fine expect one specific scenario where \r is at the end of buffer and StreamReader wants to peek the next character (to see if the next character is \n). In peek method if we have reached the end, the Peek method tries to reload the internal buffer. To see if the end is reached, the method use m_curBufPos == (m_curBufLen-1) but it does not work if the case where the end is passed.
This change fix the issue by changing just this reloading logic in Peek method. Alternatively, I could change the logic of read method. However, this change is simpler and does not have any side effect. So I prefer this fix.

smaillet-ms: Sounds like something in need of some comments explaining the situation so no-one is tempted to change things around

PR #200, opened 2015-06-23
Fix XML lib checksum and interop table

### LINKED ISSUE #168: Error in System_Xml library interop table
Due some changes to the library, the interop table needs to be updated.

mortezag:
Fixes #168


ColinA-MSFT:
:+1:

Issue #186, opened 2015-06-18
Update CMSIS support to use latest version of CMSIS


smaillet-ms:
Currently NETMF is building against CMSIS 4.2.0 where it uses CMSI. The latest revision of CMSIS is now at 4.3.0. Prior to finalizing the v4.4 release it should be building against the latest release of CMSIS.

smaillet-ms:
Fixed in #216

PR #216, opened 2015-06-27
Update CMSIS to 4.3.0

cw2:
Fix for #186. Updated RTX related settings, it is now part of the CMSIS distribution, in `RTOS\RTX` subdirectory.


DeviceCode/Targets/OS/CMSIS_RTOS/CMSIS_RTOS.settings
```diff
@@ -33,5 +33,15 @@
                Condition="!EXISTS('$(SPOCLIENT)\CMSIS\ARM.CMSIS.pdsc')"
                Text="Missing CMSIS installation at: $(SPOCLIENT)\CMSIS\*"
                />
+        <!-- Check CMSIS version -->
```
smaillet-ms: Nice extra touch cracking open the pdsc to check for the version! :smile:

cw2: :innocent: 

I made an explicit check ( == 4.3.0) due to RTX directory change. For the future releases, the check could be relaxed to ">= x.y", if there are no breaking changes (solutions that do not use RTX compile just fine, there are no breaking changes in CMSIS 4.3).

Issue #218, opened 2015-06-30
STDIO is not initialized in TinyBooter


cw2:
_Split from #217_

I've found another problem related to ITM - it does not work in Tinybooter. Respectively, it now only works when all four ports (DEBUG_TEXT_PORT, STDIO, DEBUGGER_PORT and MESSAGING_PORT) are set to the same value and it has to be either USBx or COMx.

In Tinybooter (Initialization\tinyhal.cpp) the communication initialization function CPU_InitializeCommunication() is never called (conditional compilation), so the initialization of STDIO depends on its type - if it is set to ITM0, it never happens, if it is set to USBx or COMx, it is initialized when either debugger or messaging does it, whichever is called first.

smaillet-ms:
As discussed the issue is technically unrelated to ITM. It has to do with how the startup and transport initialization and startup work in tinybooter vs. the other projects. It's not clear why that was done differently and if there is any size/perf benefits to short cutting that over keeping things consistent and standard. My current recommendation on this is to modify the tinybooter start up so it is consistent and calls CPU_InitializeCommunication().

smaillet-ms:
I'll fix this at the same time as I fix the STM32 UART code. (in progress now)

Issue #10, opened 2015-03-13
STM32F4 USART is broken


Savvkin:
In the file `STM32F4_usart_functions.cpp` in `CPU_USART_Initialize` function at [Line:112](https://github.com/NETMF/netmf-interpreter/blob/dev/DeviceCode/Targets/Native/STM32F4/DeviceCode/STM32F4_USART/STM32F4_usart_functions.cpp#L112) you decrement the `ComPortNum` value and then call `CPU_USART_GetPins` at [Line:182](https://github.com/NETMF/netmf-interpreter/blob/dev/DeviceCode/Targets/Native/STM32F4/DeviceCode/STM32F4_USART/STM32F4_usart_functions.cpp#L182).
Inside `CPU_USART_GetPins` at [Line:465](https://github.com/NETMF/netmf-interpreter/blob/dev/DeviceCode/Targets/Native/STM32F4/DeviceCode/STM32F4_USART/STM32F4_usart_functions.cpp#L465) you decrement the `ComPortNum` again.

The rest of the file contains similar errors.

smaillet-ms:
Yep, that whole virtual port for the ITM has turned out to be one of those "seemed like a good idea at the time" things... Probably the best option is just rolling that back out. The whole Communications redirector and multiple xxx_printf variants in the HAL/PAL is in need of some heavy refactoring/re-design as plugging in the ITM channel should be a brain dead simple thing.

smaillet-ms:
Fixed in #223 and #222

PR #222, opened 2015-07-01
Fix itm virtual port

smaillet-ms:
Fix for #218 and #10


smaillet-ms:
closing/recalling this due to confusion/inconsistencies on DebugTextPort vs stdio will fix and resubmit

PR #226, opened 2015-07-08
Add AllJoyn support to NETMF

mortezag:


mortezag:
I'll open another pull-request later after some modifications.

PR #243, opened 2015-07-13
Removed use of $(ProjectDir) from output and intermediate paths

### LINKED ISSUE #238: OutputPath property does not support absolute paths
http://netmf.codeplex.com/workitem/2289

smaillet-ms:
Fixes #238 
- removed all use of $(ProjectDir) in output and intermediate paths since that assumes those paths are all relative.


smaillet-ms:
:-1: This needs to account for deployment as well. Apparently the use of $(ProjectDir) was to assist the deployment process... and with these changes deployment can't find anything to deploy. Therefore closing until I can resolve that.

PR #244, opened 2015-07-13
Fix vs14 vsix solution

smaillet-ms:
- Updates the VSISX and VS integration solution for VS2015 to allow building and debugging the integration in the VS SDK experimental hive.
- removes VS Express SKU support now that Community Edition is freely available. This allows integrations without the burden of a Package Load Key and allows publishing to the official VS gallery directly (instead of as a link like we need to do for express SKUs)
- renamed the template generating projects as having multiple projects with the same assembly name was confusing the build system


smaillet-ms:
some dependencies here. Ultimately the build tasks must be contained in their own solution and manually built first. Since the tasks they build are used by the rest of the projects they are needed for their build. Unfortunately you can't rebuild them once a build has occurred as msbuild will leave a service instance of itself around for a while with all previously loaded tasks to speed up builds where the same set of tasks are loaded often. This means that once you build in VS the tasks are loaded and will fail to complete a new build since it can't overwrite the output files, as msbuild still has them loaded. Thus the tasks cannot be built from the same solution as the one building them, and when seitching to the solution that does build them killing the pending msbuild service instance may be required (It will time out after several minutes on it's own but who wants to wait for that!)

PR #245, opened 2015-07-14
STM32F4DISCOVERY: TinyBooter uses ITM0

cw2:
This is implementation of ITM virtual port fix #223 for the STM32F4DISCOVERY solution. TinyBooter diagnostic output (STDIO) is set to ITM0, USART driver and PAL replaced with stubs (now both GCC and MDK binaries fit into 32 KB).


smaillet-ms:
:+1:

PR #247, opened 2015-07-15
Replace AlwaysPrintf with Debug.Print

mortezag:


Product/Samples/AlljoynBasicService/AlljoynBasicService/AlljoynBasicService.cs
```diff
@@ -99,7 +100,7 @@ public static void Main()
                     {
                         goto exit;
                     }
-                    myAlljoyn.AlwaysPrintf(("StartService returned AJ_OK; running \n"));
```
smaillet-ms: Any reason not to remove this API entirely from the managed interop wrapper? It's misleading by name and not consistent with .NET in general and NETMF in particular (as illustrated by this change)

mortezag: Done.

smaillet-ms:
:+1:

PR #260, opened 2015-07-28
Support to build VS integration directly in VS2015

smaillet-ms:
- removed VS express SKU support as it causes issues with publishing to VS gallery and we don't need it anymore now that the VS community edition is freely available.
- fixed VS14 integration solution and projects to allow build and debug with VS experimental Hive
- added missing buildTasks project to depenencies for VS integration solution
- added missing WinUsb project reference for VS integration
- Resolved circular dependency on build task projects
- removed project references to build task projects as they are needed to build other projects
- Create standalone solution to build the build task projects needed by other projects to allow building them independently of everything else. (NOTE: that msbuild.exe will normally leave an instance of itself running with all previously loaded tasks still loaded. Thus, to build the build tasks it may be required to kill the pending msbuild.exe instance(s) to release the DLLs [It will timeout on it's own after a few minutes as well])
- Added standard project configuration settings to project files
- restored direct <Reference> tags for assemblies since the command line build from Build_sdk won't process ProjectReferences properly (Fixing that is a complete re-write of the build)
- Added some dependency logging to the Common targets file to log child projects to help in tracking dependency issues.
- updated VSIX manifest to official RTM name of Visual Studio 2015
- Fixed solution configuration settings so it doesn't mix debug and release settings on a single solution configuration.
- Fixed WinUsb and Debugger CSPROJ files to set OutDir to the proper build output location

### DIFF
--- a/Framework/CorDebug/VsPackage.cs
+++ b/Framework/CorDebug/VsPackage.cs
@@ -63,7 +63,6 @@ namespace Microsoft.SPOT.Debugger
                           , TemplateIDsVsTemplate=VsPackage.VbTemplateIds
                           )
     ]
-    [ProvideExpressLoadKey( "4.3.0.0", "Microsoft .NET Micro Framework SDK", "Microsoft Corporation", VsWinExpressId=1001 )]
     [ProvideObject( typeof( PropertyPageComObject ) )]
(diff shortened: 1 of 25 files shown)


mortezag:
:+1:

Issue #278, opened 2015-08-09
SolutionWizard no longer in tools


SytechDesigns:
Solution Wizard is missing in the MF4.4 tools. Will there be a new tool for cloning solutions or will Solution Wizard be added back to tools. At the moment I am using Solution Wizard from MF4.3 to clone solutions.

JohnySeven:
Hi, check this pull request: https://github.com/NETMF/netmf-interpreter/pull/122

SytechDesigns:
Thanks for the reference - I see, but as in #249, we do need some sort of tool for cloning solutions. For now I will use the old 4.3 version - or maybe even go back to some old power shell scripts we used to use to clone solutions. Solution Wizard could be a 'odd',  but it does give a reasonable base to edit for the new clone.

smaillet-ms:
Unfortunately, with the inclusion of the RTOS to fix multiple LWIP stack issues solution wizard is just broken beyond repair. To properly handle all the "special cases" required would mean a massive re-design. Though we'd welcome any suggestions or proposals on designs for a simpler solution to get a basic clone going.

cw2:
I've cloned a few solutions recently and as long as there is a good source, it is not that hard - a few tips to make it a little bit easier:
* Order driver libraries (\<ItemGroup>\<xxxLibs.../>\<Required Projects... />\</ItemGroup>) alphabetically in .proj file. Its ~20 lines of C# or similar and makes it really easy to copy and most importantly to compare those files.
* Similarly (a few lines) to generate/update new GUIDs (\<ProjectGuid>),
* Replace \<ProjectPath> with $(MSBuildThisFileFullPath) to remove hardcoded path,
* Delete unused elements that contain solution name - like \<CustomSpecific>

IMHO those two last items applied en-masse would reduce the cloning work to a minimum, most of the build variables (like Directory, various Solution-base ones etc.) could be either defined relative to the project file or based on (one) value in the main .proj or .settings file. It should be enough to define only the solution name.

lt72:
Cleanup is definitely something we should be doing in the background and I agree that it can go a long way. I also think that having a white paper could be helpful... or a wiki.

cw2:
Well, I have a few things already on my list (that started long time ago), which I am more than willing to contribute - I guess I could start in a dedicated 'cleanup' branch with small/simple deletes in mostly random areas and discuss bigger tasks in some 'Cleanup' topic/issue... (?)

PR #283, opened 2015-08-15
Cleanup for cloning

cw2:
As discussed in #278, deleted unused \<CustomSpecific> tags, empty ProjectPath attributes and unused \<ProjectPath> tags from solution projects files.

As it turned out, \<ProjectPath> is only referenced from CLR\Tools\PlatformDesigner\*, which is not used anymore, so I believe \<ProjectPath> can be safely removed (from the rest of files, along with ProjectPath attribute) ?

### DIFF
--- a/Application/MicroBooter/MicroBooter.proj
+++ b/Application/MicroBooter/MicroBooter.proj
@@ -13,8 +13,6 @@
     <Groups>MicroBooter</Groups>
     <Documentation>
     </Documentation>
-    <CustomSpecific>
-    </CustomSpecific>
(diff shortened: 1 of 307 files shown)


cw2:
The reason of deletion is to reduce number of places where solution name or directory has to be changed during cloning.

smaillet-ms:
FYI: we haven't forgotten this, we are holding pull requests that aren't bug fixes critical to the v4.4 release until we get that release out. Once it is out (shortly) we'll come back to all pending pull requests

Issue #56, opened 2015-03-31
StringTable update and cleanup


cw2:
The string table (`c_CLR_StringTable_Data[]`, `c_CLR_StringTable_Lookup[]`) in `CLR\Core\StringTable.cpp` was generated from assemblies that do not exist anymore and its content is outdated (~20%, cca 3 KB). Also, the mechanism used to create it has changed and there is some obsolete code and unused files.

So, I'd suggest to
1. Modify `MetaDataProcessor.exe` to exclude strings such as `$$method0x6000*` and `<PrivateImplementationDetails>{GUID}` from generated string table,
2. Update list of `StringTableInputs` in `Microsoft.SPOT.Support.Settings` to include valid *.strings,
3. Generate new string table and update `StringTable.cpp` content,
4. Delete obsolete `tools\scripts\compile_stringtable.cmd`,
5. Delete code in `#if defined TINYCLR_CREATE_STRINGTABLE`.

cw2:
Perhaps a better alternative to 1. (+ 3.) is to create an inline msbuild task directly in `Microsoft.SPOT.Support.Settings` (?)

cw2:
I volunteer for this issue, I guess there is a lot of more important things you have to do...

lt72:
Metadataprocessor generates the string table and it happens through the -loadStrings and -generateStringsTable switches.  The obsolete script  should be updated with a fresh list of assemblies but having a dedicated task would actually be ideal.
Whatever we do, the important bit is to document and log clearly, before or after the build process, in a script or in a task, which assemblies are we including in the StirngTable generation.
Last, the more assemblies one includes, the more strings you get. More strings, faster perf but also larger code size.

example of config script the obsolete script depends on below: 


Cmd>more tools\Libraries\opt_stringtable.cfg
 -loadStrings %BUILD_TREE_CLIENT%\pe\mscorlib.strings
 -loadStrings %BUILD_TREE_CLIENT%\pe\Microsoft.DirectBand.Protocol.strings
 -loadStrings %BUILD_TREE_CLIENT%\pe\Microsoft.SPOT.Application.Channels.strings
 -loadStrings %BUILD_TREE_CLIENT%\pe\Microsoft.SPOT.Application.Data.strings
#-loadStrings %BUILD_TREE_CLIENT%\pe\Microsoft.SPOT.Application.Samples.strings
[...]

PR #284, opened 2015-08-17
Issue 56 - StringTable update and cleanup

cw2:
Fix for #56.

The StringTable is now generated via `buildstring` target:
```
msbuild Tools\Targets\Microsoft.SPOT.Support.Settings /t:buildstring
```
the output `BuildOutput\public\Release\Client\stubs\StringTableData.cpp` can directly replace `CLR\Core\StringTableData.cpp`.

The contents is based on `Microsoft.SPOT.Native.strings` and `mscorlib.strings`, the only valid entries in the original inputs. Other strings could be added in future revisions, most likely Microsoft.SPOT.Hardware.Usb and Microsoft.SPOT.Hardware.SerialPort, which are often included in the firmware image (perhaps after some code size measurements).


mortezag:
:+1:

PR #287, opened 2015-08-24
Fix UTF8 to UTF16 conversion for special characters

### LINKED ISSUE #131: Data corruption issue with UTF16 surrogate pairs
I'm running into a data corruption issue on NETMF v4.3 when dealing with UTF16 surrogate pairs inside of char arrays. The following code demonstrates the issue I'm seeing:

```cs
Debug.EnableGCMessages(false);
for (int i = 0; i < 10; i++)
{
    string s = string.Empty;
    char[] buffer = new char[2];
    for (int j = 0; j < 30; j++)
    {
        buffer[0] = (char)0xD83D;
        Debug.GC(true);
        buffer[1] = (char)0xDE80;
        s += new string(buffer);
    }
    Debug.Print("Result: " + s);
}
```

The character I'm attempting to build in this example is the rocket emoji `"\uD83D\uDE80"`. The issue I'm seeing is that random characters are dropped or added and, occasionally, parts of other strings (and method names, etc.) are pulled in to the final result.

Here's an example of the output I'm seeing:
utf16issue-netmf

Here's what it should look like (the same code run from a desktop console app):
utf16issue-desktop

I've been able to observe this issue on both the NETMF emulator and on a Netduino 3 Wi-Fi.

This issue actually appears pretty frequently when dealing with StringBuilders, etc. I'm not sure what specifically is triggering the issue, but I have noticed running the garbage collector makes it appear more frequently. Even without the call to Debug.GC in the above code, the result will have some corruption (although usually not as much).

Please let me know if you need any more information. Thanks!

mortezag:
Fixes #131

### DIFF
--- a/CLR/Core/CLR_RT_UnicodeHelper.cpp
+++ b/CLR/Core/CLR_RT_UnicodeHelper.cpp
@@ -103,6 +103,7 @@ int CLR_RT_UnicodeHelper::CountNumberOfBytes( int max )
             if(ch >= LOW_SURROGATE_START && ch <= LOW_SURROGATE_END)
             {
                 num += 4;
+                max--;
             }
             else
             {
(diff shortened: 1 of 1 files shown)


smaillet-ms:
:+1:

PR #291, opened 2015-08-31
Fixed MCBSTM32F400 networking build with latest CMSIS-RTX

smaillet-ms:
- updated cmsis_os_cpp.h to latest RTX data structure layout
- updated C++ support macros for timeres and mutex to maintain consistency between them for use. (e.g. 3 seperate macros, one to declare a member of a class, one for the constructor initializaion list and a third to be used in the constructor to perform final init/construction. This pattern ensures that the correct action is taken and allows for greater version implementation detal isolation. It is possible that the internals of the macro can change or even be empty in some cases and then in a new version change.
- forced all RAM regions propr to the LWIP init region to uncompressed, this is to work around the linker bug in the ARMCC tool chain. WHile that bug was fixed in the latest toolchain, not everyone has upgraded. In the next release of NETMF we'll reset the min requiremeents to include the fix - until then we'll keep the workaround.
- fixed version check in cmsis_os_cpp.h that should have caught these issues immediately but didn't; the check is now for an exact match as there is no guarantee of upwards compatibility


Solutions/MCBSTM32F400/netmfdbg/netmfdbg.uvoptx
```diff
@@ -192,7 +192,56 @@
           <Name>UL2CM3(-S0 -C0 -P0 -FD20000000 -FC1000 -FN1 -FF0STM32F4xx_1024 -FS08000000 -FL0100000 -FP0($$Device:STM32F407IGHx$CMSIS\Flash\STM32F4xx_1024.FLM))</Name>
         </SetRegEntry>
       </TargetDriverDllRegistry>
-      <Breakpoint/>
+      <Breakpoint>
```
lt72: these seems to be an unnecessary local config file pointing to a local HD that should be removed

Solutions/MCBSTM32F400/netmfdbg/netmfdbg.uvoptx
```diff
@@ -710,38 +815,38 @@
       <Books>
         <Book>
           <Number>0</Number>
-          <Title>Schematics (MCBSTM32F400)</Title>
-          <Path>C:\Keil\ARM\PACK\Keil\STM32F4xx_DFP\2.2.0\MDK\Boards\Keil\MCBSTM32F400\Documentation\mcbstm32f400-schematics.pdf</Path>
+          <Title>User Manual (MCBSTM32F400)</Title>
+          <Path>C:\Keil_v5\ARM\PACK\Keil\STM32F4xx_DFP\2.5.0\MDK/Boards/Keil/MCBSTM32F400/Documentation/mcbstm32f200.chm</Path>
```
lt72: other local config?

Solutions/MCBSTM32F400/netmfdbg/netmfdbg.uvoptx
```diff
@@ -1078,6 +1281,260 @@
     </TargetOption>
   </Target>
 
+  <Target>
```
lt72: Quite a bit of changes. Is this driven by the tool version change?

Solutions/MCBSTM32F400/TinyCLR/scatterfile_tinyclr_mdk.xml
```diff
@@ -114,7 +114,7 @@
             <FileMapping Name="*" Options="(VectorTable +FIRST)" />
         </ExecRegion>
         
-        <ExecRegion Name="ER_RAM_RW" Base="+0" Options="ABSOLUTE" Size="">
+        <ExecRegion Name="ER_RAM_RW" Base="+0" Options="ABSOLUTE NOCOMPRESS" Size="">
```
lt72: is 'NOCOMPRESS' a new keyword and is that necessary?

Solutions/MCBSTM32F400/TinyCLR/scatterfile_tinyclr_mdk.xml
```diff
@@ -114,7 +114,7 @@
             <FileMapping Name="*" Options="(VectorTable +FIRST)" />
         </ExecRegion>
         
-        <ExecRegion Name="ER_RAM_RW" Base="+0" Options="ABSOLUTE" Size="">
+        <ExecRegion Name="ER_RAM_RW" Base="+0" Options="ABSOLUTE NOCOMPRESS" Size="">
             <FileMapping Name="*" Options="(+RW-DATA, +ZI)" />
         </ExecRegion>
```
lt72: What about GCC?

Solutions/MCBSTM32F400/netmfdbg/netmfdbg.uvoptx
smaillet-ms: That's how uVision does it.For the most part the paths are hints and it can resolve things, but we can't really control what it stores in these files and their contents are completely undocumented.

Solutions/MCBSTM32F400/netmfdbg/netmfdbg.uvoptx
smaillet-ms: No it's driven by uVision configuration to the project to debug the "nonet" project as welll

Solutions/MCBSTM32F400/TinyCLR/scatterfile_tinyclr_mdk.xml
smaillet-ms: New-ish, and yes it is necessary, As stated in the PR comments this gets around a linker bug for the ARM linker. While the bug is fixed in the latest released version not everyone has migrated to that version and this works around the problem. When we get to the v4.5 release we can require the latest linker as a minimum requirement and verify everything functions properly.
Without this any images built using the older linker will fail to debug apps from VS as the restart to reconnect after deploy will fail triggering a hard fault in the processor.

Solutions/MCBSTM32F400/TinyCLR/scatterfile_tinyclr_mdk.xml
smaillet-ms: This issue is ARM linker specific and not applicable to GCC.

lt72:
:+1:

PR #295, opened 2015-09-09
Restored Legacy Crypto support

smaillet-ms:
- adding in support for legacy crypto, which was removed previously. Turns out it is still needed by TinyBooter/mfdeploy for firmware signature checks as a small footprint crypto lib.

**NOTE:** This will fail to build the SDK without the actual binary DLLs. These were available as an additional MSI installer for the old Porting kit. We'll have an updated MSI containing the binaries soon. Longer term we'll be looking at how to completely remove these libraries.


mortezag:
:+1:

PR #298, opened 2015-09-14
Adding support to provide more useful and actionalble diagnostics for Crypto libs and wire protocol debugging

smaillet-ms:
- Added code to tinybooter to dump out newkey, signature and current key on failure to verify signature on key updates to aid in diagnosing issues with Crypto library.
- fixed multiple issues of return or throw on smae line as conditional expression which make debugging difficult. (Some debuggers can't set breakpoints on line and column)
- for the crypto.lib project forced mapping MDK to the RVDS3.1 thumb2 crypto libraries.
- added more readable wire-protocal packet tracing, now prints the command id as a textual name along with a decoded form of the flags with textual names for all active flags to aid in debugging the communications.
- fixed DLLImports for the Crypto.dll to use CDecl calling convention to prevent issues with stack corruption and eliminate MDA exception while debugging.
- Fixed flash driver for MCBSTM32F400 to check for ready status after erase to ensure chip is in a valid state before continuing. Othereise the chip can get corrupted (particularly the Config sector)
- Added Addditional debug and ETW tracing for the NETMF wireprotocol in MFDeploy and VS integration.
- Added additional trace messaging support toe device side wire protocol to help in tracking any failures that might occur on a device.

### DIFF
--- a/Application/TinyBooter/Commands.cpp
+++ b/Application/TinyBooter/Commands.cpp
@@ -1452,6 +1452,17 @@ bool Loader_Engine::Monitor_CheckSignature( WP_Message* msg )
     return true;
 }
 
+#ifdef DEBUG
+// dumps binary block in a form useable as C code constants for isolated testing and verification
+void DumpBlockDeclaration( char const* name, UINT8 const* pBlock, size_t len )
+{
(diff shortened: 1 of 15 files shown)


mortezag:
:+1:

PR #300, opened 2015-09-15
Enabling time service for MCBSTM32F400

smaillet-ms:
NOTE: This is for the fixCrypto branch


CLR/Core/Execution.cpp
```diff
@@ -840,7 +840,7 @@ bool CLR_RT_ExecutionEngine::SpawnStaticConstructorHelper( CLR_RT_AppDomain* app
     CLR_RT_MethodDef_Index idxNext;
 
     _ASSERTE(m_cctorThread != NULL);
-    _ASSERTE(m_cctorThread->CanThreadBeReused());  
+    //_ASSERTE(m_cctorThread->CanThreadBeReused());
```
smaillet-ms: This assert is benign and will trigger sometimes in debug builds so it is being removed.

mortezag:
:+1:

PR #301, opened 2015-09-16
ENabled debug info for release and RTM builds

smaillet-ms:
- Enabled debug symbol information for release and RTM builds with MDK toolchain (It doesn't impact optimizations but allows for actual source debugging of release builds via JTAG debugger)


mortezag:
:+1:

PR #303, opened 2015-09-21
Fixed SSL support for v4.4

smaillet-ms:
-Removed WSAEWOULDBLOCK as it's existence causes conditional checks to generate incorrect code for NETMF devices. (basically if WASEWOULDBLOCK exists AND is the same as EWOULDBLOCK on a non-windows targeted build, then EWOULDBLOCK gets treated as a fatal error instead of a bening notification) In previous NETMF versions the modified LWIP port handled the would block case differently, such that this error was never seen. In moving the LWIP stack to another thread this issue was discovered.


lt72:
:+1:

PR #304, opened 2015-09-22
Added HttpClient Sample.

smaillet-ms:
NOTE: The sample does not build with and is not included in the SDK. Rather the intention is that we will publish samples for NETMF in the Visual Studio Gallery. This allows updating and adding samples without requireing a new SDK. It also enables uses to find additional third party samples from a central place.


lt72:
:+1:

PR #309, opened 2015-09-25
Enc28j60

techcap:
changed lwip path to lwip_1_4_1_os.

On enc28j60 driver, GLOBAL_LOCK interferes with lwip memory functions such as pbuf_alloc.
So before using memory functions, isr disable should be restored.


Test/Platform/Tests/CLR/System/Text/Regexp/GroupTests.cs
```diff
@@ -1,4 +1,4 @@
-using System;
```
smaillet-ms: This file is unrelated to ENC28J60, please remove it from this pull request.

techcap: This file is not related to the enc28j60 driver.
But GroupTests.cs file has ANSI encoding and TM special character so in non-western, it occurs compile problem. I changed it to Unicode. Should I have to separate it?

smaillet-ms: Yes, this is an unrelated issue and should have a seperate bug filed.

PR #311, opened 2015-09-25
Change GroupTests.cs encoding from ASCII to Unicode

techcap:
Change GroupTests.cs encoding from ASCII to Unicode


mortezag:
:+1:

PR #313, opened 2015-09-25
STM32F4x9_Flash_driver.cpp also needs FLASH definition

techcap:
STM32F4x9_Flash_driver.cpp also needs FLASH definition.
So I moved FLASH definition to the header file.


mortezag:
:+1:

PR #315, opened 2015-09-25
Fix TimerServiceTests

mortezag:
Replace hardcoded ip address for time server with a more robust code which gets the ip from time.nist.gov server.


smaillet-ms:
:+1:

PR #317, opened 2015-09-29
Set error on LINGER and DONTLINGER socket options

mortezag:
Set error for both LINGER and DONTLINGER options as neither of them are supported by the current LWIP


smaillet-ms:
:+1:

PR #322, opened 2015-10-02
Fix NetInfo Tests

mortezag:
- Modify NetworkInterfaceTest4_DynamicIP_StaticDNS and NetworkInterfaceTest6_StaticIP_DynamicDNS as these combinations are not supported at this moment
- Comment out NetworkInterfaceTest8_GetHostEntry since it is failing and its intention is not clear


lt72:
:+1:

PR #323, opened 2015-10-06
Fix http tests issues

mortezag:
- Use a utility method to return local host address instead of using 127.0.0.1
- Remove WebProxy from http requests
- A few other minor fixes


lt72:
:+1:

PR #324, opened 2015-10-06
Revert "STM32F4x9_Flash_driver.cpp also needs FLASH definition"

mortezag:
Reverts NETMF/netmf-interpreter#313

### DIFF
--- a/DeviceCode/Targets/Native/STM32F4/DeviceCode/STM32F4_Flash/STM32F4_Flash.h
+++ b/DeviceCode/Targets/Native/STM32F4/DeviceCode/STM32F4_Flash/STM32F4_Flash.h
@@ -19,10 +19,6 @@
 
 //--//
 
-#ifndef FLASH
-#define FLASH               ((FLASH_TypeDef *) FLASH_R_BASE)
-#endif
-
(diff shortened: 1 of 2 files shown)


smaillet-ms:
This was reverted as it isn't something that should be done this way. In order to prevent confusion and frustration - the override of the "FLASH" macro should be done in a local a scope as possible (e.g. one single compilation unit) That macro has meaning elsewhere that is distinct from the use here, which is why it had to be overridden. However, putting the override into a header file means that the door is open to accidentally bringing it into other compilation units. Worse, if the ordering of include files is changed it is possible that the "override" is actually overridden by the original definition. These sorts of issues are maddeningly difficult to chase down. So the override should stay to as narrow a scope as possible. The only reason for using it at all is to resolve the definitions established by two independently created code bases brought together in NETMF.

techcap:
So, finally you mean the definition of FLASH macro should be placed in both STM32F4_Flash_driver.cpp and STM32F4x9_Flash_driver.cpp. Right?

smaillet-ms:
Yes, the goal is to reduce the scope to as small an area as possible to avoid any unintended consequences of an overloaded macro definition.

PR #328, opened 2015-10-13
Update Certificate Ids for official signing

smaillet-ms:
- Added git ignore for the binary only crypto libraries so they don'tget added into the tree
- Updated codesign certificate ids to current Microsoft SHA2 variants.


mortezag:
:+1:

Issue #330, opened 2015-10-20
fix for emulator so it doesn't block when there is no data on a SSL connection


josesimoes:
This is fix that as been proposed by Xin Chen of the AMQP Lite project.
See:
http://amqpnetlite.codeplex.com/wikipage?title=NETMF%20and%20TLS%2fSSL%20connection&referringTitle=Documentation

smaillet-ms:
Please submit the change as a pull request so it can be reviewed and merged into the main code repository.

josesimoes:
Done! PR #342

PR #342, opened 2015-10-29
Fix emulator hanging when using TLS/SSL

josesimoes:
Implement fix suggested in #330

This is a fix that has been proposed by Xin Chen of the AMQP Lite project.
See:
http://amqpnetlite.codeplex.com/wikipage?title=NETMF%20and%20TLS%2fSSL%20connection&referringTitle=Documentation


Framework/Tools/Emulator/Ssl.cs
```diff
@@ -608,9 +650,13 @@ public int DataAvailable( int socket )
 
             if (!GetSslData(socket, out ssd)) return (int)SocketError.SocketError;
 
-            if(!_socketsDriver.GetSocket(socket, out sock)) return (int)SocketError.SocketError;
+            if (!_socketsDriver.GetSocket(socket, out sock)) return (int)SocketError.SocketError;
```
smaillet-ms: Please split this line, never put the return on the same line (While there's a ton of legacy code in NETMF some of which has this practice we are trying to clean it up so let's not add any new code following that style)

Framework/Tools/Emulator/Ssl.cs
```diff
@@ -608,9 +650,13 @@ public int DataAvailable( int socket )
 
             if (!GetSslData(socket, out ssd)) return (int)SocketError.SocketError;
 
-            if(!_socketsDriver.GetSocket(socket, out sock)) return (int)SocketError.SocketError;
+            if (!_socketsDriver.GetSocket(socket, out sock)) return (int)SocketError.SocketError;
 
-            return sock.Poll(0,SelectMode.SelectRead) ? 1024 : 0;
+            var readData = ssd.m_readData;
+            int ret = readData != null ?
```
smaillet-ms: Please split this up into If/else it's pretty dense and anything that forces the reader to have to stop and think, just to parse the code let alone understand its intent is something to avoid.

Framework/Tools/Emulator/Ssl.cs
josesimoes: Done!

Framework/Tools/Emulator/Ssl.cs
josesimoes: Done too!
I agree with you 100%. 
Remember that I'm not the author of these changes, I just applied the patch coming from the AMQP Lite project.

PR #343, opened 2015-11-04
Update for BitConverter that avoids reinterpret cast on unaligned memory

### LINKED ISSUE #306: Running BitConverter tests crashes
Running BitConverter tests on MCBSTM32F400 (Keil) crashes and make the board not responsive. I have to erase the deployment region to make it work again.

doingnz:
reinterpret_cast<T>(v) is in general dangerous. It can inadvertently attempt to access unaligned memory when it tries to cast a BYTE\* to something else. To avoid this risk, this update creates a local variable of the required type, then copies the byte aligned values byte by byte into this local variable before using the local variable to set the return value on the stack. Used the same code layout for all for readability. Assuming compiler will optimize sufficiently. 

The existing Unit Tests will all pass when run on iMXS system after applying this update.   

Hopefully fixes #306 

Note the pull also includes changes for adding ARM compiler DS5. I am not sure how to create a pull request that only includes the BitConverter change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,10 @@ Listings/
 /DeviceCode/Targets/OS/Win32/DeviceCode/WinPcap_Eth/Dependencies/WpdPack/
 
 *.axfdump
+/Test/Platform/Tests/CLR/Microsoft.SPOT.TinyCore/Presentation/OnBoardFlash.dat
+/Test/Platform/Tests/CLR/Microsoft.SPOT.TinyCore/Presentation/OnBoardFlash.dat.smd
+/Test/Platform/Tools/MFTestSystem/Results
+/devenvdebug.cmd
+/setenv.cmd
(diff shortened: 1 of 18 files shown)


smaillet-ms:
Please squash those commits down to a single commit (There's a lot of merging going on that's not historically relevant to the actual changes made)

smaillet-ms:
Also, please split out the DS5 support. Pull requests should focus on single fix/feature so they are easier to understand and if needed, roll back.

smaillet-ms:
@doingnz any chance of updating this PR as requested in comments? Otherwise it will be closed without merging into the main tree.

uscomrs:
I have created a new Pull Request #395 with the changes based on the latest 'dev' branch.

smaillet-ms:
Thanks for the update!

PR #359, opened 2015-12-10
STM32F4 RTC support

josesimoes:
- add support for hardware RTC of STM32F4
- configuration follows available hardware options: 
  - internal RC
  - external low power oscillator
  - external low power with oscillator bypass
- the configuration is performed in platform_selector.h and tinybooter/tinyCLR proj files
- one of the important changes is that (providing the RTC pin is properly powered) the mcu doesn't loose time between reboots (actually the boot code won't zero the RTC if there is a valid value in the RTC registers)
- this was tested on a STM32F4 Discovery board fitted with a 32.768kHz crystal

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,8 @@ Listings/
 
 *.axfdump
 /crypto/lib
+Firmware License (Crypto).rtf
+License (Crypto).rtf
+PKCryptoWelcome.rtf
+REDIST_CRYPTO.TXT
+ReleaseNotesCrypto.txt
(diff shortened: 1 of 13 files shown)


mbyamukama:
Thank you! Could you provide sample code for this? I am particularly interested in RTC interrupts as they are helpful in waking up the device from DeepSleep. Are they supported? Using the internal RC (native programming using C) sometimes causes the time to gradually drift off, sometimes upto 2s in a minute. Have you experienced this?

PR #370, opened 2016-01-12
Fix for signing empty files

josesimoes:
When signing empty files (0 lenght) Cmd_SignFile breaks. This happens because nothing is loaded from the file and latter when the code is cheking for the buffer size it breaks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,8 @@ Listings/
 
 *.axfdump
 /crypto/lib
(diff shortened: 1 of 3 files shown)


kmshim:
@josesimoes 
I have succeeded to build STM32F4DISCOVERY solution by using your modification.
thank for your help.

josesimoes:
@kmshim :+1:

PR #386, opened 2016-02-16
Update CMSIS to 4.5

josesimoes:


smaillet-ms:
What platforms did you test this against?

josesimoes:
Discovery4

smaillet-ms:
This requires more testing before it can be merged in. At the very least it must be validated against the MCBSTM32F400 using networking, as any change in the CMSIS-RTX implementation could impact the networking sys-arch layer. (we have some dependencies on internal implementation details that require testing - and have a history of breaking things)

josesimoes:
Can't help you with that... I don't have a MCBSTM32F400 board.
I just build it for the Discovery Solution and loaded the image. The boot was OK and the board showed on MFDeploy.

This "update" on the CMSIS was made as part of my quest to add networking with GCC.

If this is not the time to update CMSIS then leave it as it is. It hasn't changed anything on networking with GCC anyway...

PR #392, opened 2016-02-17
Add extension to image files build with GCC

josesimoes:
Tthis makes the image files ready to be uploaded with ST-Link app, DFU manager or similar.


smaillet-ms:
Please squash your commits before posting a pull request so that the pull request contains only actual code change commits (e.g. no merges or reverts, etc...)

smaillet-ms:
In the future please submit an issue describing the problem and any proposed solution before submitting a pull request so that the issue and proposal may be discussed. I'm not particularly comfortable with this change. While it seems simple (just adding an extension) it has upstream implications that need to be considered. The build has generated these files without extensions (for good or ill) for the entire lifetime of NETMF. There are production systems with automated builds in place that expect the names to match and changing the names would break those. Adding the .bin can be a custom stage of the build for your case or, we could add a copy task to generate a copy of the files with the .bin extensions so we don't break anything that depends on the way it's always been done.

smaillet-ms:
In order to ensure that pull requests remain focused and easily rolled back if issues arrive please keep pull requests focused to exactly one issue. This change includes the CMSIS4.5 as well which should be a separate item unrelated to this one.

josesimoes:
@smaillet-ms undestood on both.
For the squash: I'm have a hard time with Git and its commands. I often screw up the commits and mix stuff between branches. Still learning my way... I do apologise for this.

Didn't realize that this change could have such heavy impact on all that. You are probably right...
I just got tired (for the 1M time) to have to add the extension to those image files so ST Link could take them and upload those to the flash.

Please disregard this PR.

PR #394, opened 2016-02-20
VisualBasic.NET Build and Debug in VS2015

### LINKED ISSUE #387: Creating project for VB.NET fails because of missing dependecies
# see email thread for additional information

smaillet-ms:
Fixes #387 
- updated VisualBasic Targets file to set required properties
- Updated VB project templates to disable optimization for debug builds so binaries are debuggable
- Updated the VSProject system tp check for explicit mscorlib assembly reference and if not present to add it to the internal list of assemblies for deployment.
- Added use of FLAVOR_SDK environment variable to build_sdk.cmd so that the SDK components can be built for debug configurations. (This helps when debugging the VS integration and MFDeploy)

### DIFF
--- a/Framework/CorDebug/VsProject.cs
+++ b/Framework/CorDebug/VsProject.cs
@@ -1,5 +1,6 @@
 using System;
 using System.IO;
+using System.Linq;
 using System.Diagnostics;
 using CorDebugInterop;
 using System.Collections;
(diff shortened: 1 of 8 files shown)


Framework/CorDebug/VsProject.cs
```diff
@@ -142,11 +143,24 @@ public string[] GetDependencies( bool fIncludeStartProgram, bool fPE, bool fTarg
                     AddDependencies( table, new string[] { this.m_startProgam } );
                     AddDependencies( table, this.m_startProgramReferences );
                 }
+                // if mscorlib isn't in the list of assemblies, add it as it's always required
+                // the build system in some versions of VS+MSBuild will assume that, while other
+                // versions will explicitly reference it with the InProject item property set to
+                // false so it won't show in the IDE. By testing for it and adding it if not found
+                // both cases are covered. 
+                var q = from path in table.OfType<string>( )
+                        where 0 == string.Compare( Path.GetFileName( path ), "mscorlib.dll", StringComparison.InvariantCultureIgnoreCase )
```
miloush: I believe this should be [OrdinalIgnoreCase](https://msdn.microsoft.com/en-us/library/ms973919.aspx) comparison, invariant culture considers things equal that file system does not. And you can use `string.Equals(string, string, StringComparison)` for better readability.

build_sdk.cmd
```diff
@@ -9,6 +9,8 @@ if "%1" == "-?" goto :ShowUsage
 if /i "%1" == "/h" goto :ShowUsage
 if /i "%1" == "-h" goto :ShowUsage
 
+if "%FLAVOR_SDK%" == "" FLAVOR_SDK=RTM
```
miloush: This probably was meant to be **set** FLAVOR_SDK=RTM

miloush:
:+1: works for me

PR #395, opened 2016-02-21
Change native BitConverter to not use Reinterpret Cast. 

doingnz:
This is a replacement for #343.

Change native BitConverter so that it does not use Reinterpret Cast. 
This avoids non-aligned memory access errors.


smaillet-ms:
:+1:

smaillet-ms:
NOTE: Since this is a change to the runtime itself, this is listed in the v4.5 runtime milestone.
At this point we don't have a means to distinguish between API compatible bug fixes and API changes in the versioning of NETMF. v4.5 needs to sort out the versioning story. (Ideally following Semantic versioning so it is easy to identify compatible vs. breaking changes)

PR #396, opened 2016-02-21
Add quotes to buildhelper paths in Microsoft.SPOT.System.Targets

doingnz:
The build files need to include quotes around paths to correctly handle folder names with permitted non-alpha-numeric characters (e.g. SPACE, '+',...)


smaillet-ms:
:+1:

smaillet-ms:
FYI This PR will be reverted as it breaks the MCBSTM32F400 builds. Please correct, test and re-submit it

smaillet-ms:
The problem with this PR turns out to be when one of the three regions (CompressImageFlash, CompressImageDat, CompressImageCfg) is empty. This translates to white space without the quites, however once the quotes are added it translates to copy /b "" + "a real file name" + " Another real name", which fails. Ultimately this whole task should be replaced by a simpler to read and follow inline C# task that can properly handle the blank names and any embedded spaces without breaking a sweat.

Issue #279, opened 2015-08-10
Undefined Environment Variables FLAVOR_WIN?


doingnz:
_FLAVOR_WIN_ appears to be undefined? or did I miss something? 
Do we need add back in the lines to set the default FLAVOR_WIN to init.cmd?

tools\scripts\init.cmd script in the past defined two environment variables FLAVOR_DAT and FLAVOR_WIN as follows:

```
if /I "%FLAVOR_DAT%"   == "Debug"   set FLAVOR_DAT=Debug
if /I "%FLAVOR_DAT%"   == "Release" set FLAVOR_DAT=Release
if    "%FLAVOR_DAT%"   == ""        set FLAVOR_DAT=Release

if /I "%FLAVOR_WIN%"   == "Debug"   set FLAVOR_WIN=Debug
if /I "%FLAVOR_WIN%"   == "Release" set FLAVOR_WIN=Release
if    "%FLAVOR_WIN%"   == ""        set FLAVOR_WIN=Release

```

In init.cmd in MF v4.4, the above lines defining a default for FLAVOR_WIN have been removed leaving FLAVOUR_DAT alone.

FLAVOR_WIN is used later in the same init.cmd and it has not been defined.

```
set BUILD_TREE_SERVER=%BUILD_ROOT%\%FLAVOR_WIN%\server
```

also in setenv_base.cmd
```
set PATH=%CLRROOT%\BuildOutput\Public\%FLAVOR_WIN%\Test\Server\dll;%PATH%
```

It is referenced in many other project files..

```
  C:\Projects\BPplus\netmf\netmf-interpreter\BuildEnv.props(23):    <FLAVOR_WIN Condition="'$(FLAVOR_WIN)'=='' AND '$(Configuration)'=='Debug'" >Debug</FLAVOR_WIN>
  C:\Projects\BPplus\netmf\netmf-interpreter\BuildEnv.props(24):    <FLAVOR_WIN Condition="'$(FLAVOR_WIN)'=='' AND '$(Configuration)'!='Debug'" >Release</FLAVOR_WIN>
  C:\Projects\BPplus\netmf\netmf-interpreter\setenv_base.cmd(74):set PATH=%CLRROOT%\BuildOutput\Public\%FLAVOR_WIN%\Test\Server\dll;%PATH%
[...]

PR #399, opened 2016-02-23
tools/scripts/init.cmd to set FLAVOR_WIN if not set

doingnz:
This pull request is to close out #279 by adding back script to set FLAVOR_WIN.

It follows the same logic as setting FLAVOR_DAT, with the exception that it will default to the value of FLAVOR_DAT if it is not defined. (FLAVOR_DAT defaults to Release if not defined)

Other minor changes to the case names of paths to match the values in BuildEnv.props

### DIFF
--- a/tools/scripts/init.cmd
+++ b/tools/scripts/init.cmd
@@ -16,6 +16,10 @@ if /I "%FLAVOR_DAT%"   == "Debug"   set FLAVOR_DAT=Debug
 if /I "%FLAVOR_DAT%"   == "Release" set FLAVOR_DAT=Release
 if    "%FLAVOR_DAT%"   == ""        set FLAVOR_DAT=Release
 
+if /I "%FLAVOR_WIN%"   == "Debug"   set FLAVOR_WIN=Debug
+if /I "%FLAVOR_WIN%"   == "Release" set FLAVOR_WIN=Release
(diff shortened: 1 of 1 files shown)


tools/scripts/init.cmd
```diff
@@ -16,6 +16,10 @@ if /I "%FLAVOR_DAT%"   == "Debug"   set FLAVOR_DAT=Debug
 if /I "%FLAVOR_DAT%"   == "Release" set FLAVOR_DAT=Release
 if    "%FLAVOR_DAT%"   == ""        set FLAVOR_DAT=Release
 
+if /I "%FLAVOR_WIN%"   == "Debug"   set FLAVOR_WIN=Debug
+if /I "%FLAVOR_WIN%"   == "Release" set FLAVOR_WIN=Release
+if    "%FLAVOR_WIN%"   == ""        set FLAVOR_WIN=%FLAVOR_DAT%
```
smaillet-ms: Default should be Release NOT FLAVOR_DAT. FLAVOR_DAT is an unrelated setting and, if we change the default for that in the future it shouldn't bleed over into FLAVOR_WIN.

doingnz: I have updated the branch to default FLAVOR_WIN to Release. :-)

smaillet-ms:
:+1:

PR #273, opened 2015-08-05
Adding support for ARM DS5 tool chain. 

doingnz:
I need to add support for ARM DS5 or RVDS 4.1

Preference is for ARM DS5 as this is in theory the same compiler (5.05) as included in MDK and is the tool chain supported by ARM going forward. There are some differences in the install paths and folders for the ARM DS5 tools, library, and include files compared to what was setup for MDK.. 

The following files are based on the MDK files and edited to remove legacy MDK versions prior to 5.05, and accommodate the folder names etc. 

They assume 
```
COMPILER_TOOL= DS5
COMPILER_TOOL_VERSION_NUM=5.05
```

	modified:   setenv_base.cmd
	new file:   setenv_ds5.cmd
	new file:   tools/Targets/Microsoft.Spot.system.ds5.settings
	new file:   tools/Targets/Microsoft.Spot.system.ds5.targets

If I try build either MCBSTM32F400 or STM32F4DISCOVERY then I get the following errors.
```
Target "ArmCompileC" in file "C:\Projects\BP+\netmf\netmf-interpreter\tools\targets\Microsoft.SPOT.System.DS5.targets" from project "C:\Projects\BP+\netmf\netmf-interpreter\DeviceCode\Cores\arm\Processors\CortexMx\ItmPort\dotNetMF.proj" (target "ArmBuildLib" depends on it):
Skipping target "ArmCompileC" because it has no outputs.
Though the target has declared its outputs, the output specification only references empty properties and/or empty item lists.
Done building target "ArmCompileC" in project "dotNetMF.proj".
Target "ArmCompileCPP" in file "C:\Projects\BP+\netmf\netmf-interpreter\tools\targets\Microsoft.SPOT.System.DS5.targets" from project "C:\Projects\BP+\netmf\netmf-interpreter\DeviceCode\Cores\arm\Processors\CortexMx\ItmPort\dotNetMF.proj" (target "ArmBuildLib" depends on it):
Building target "ArmCompileCPP" completely.
Output file "C:\Projects\BP+\netmf\netmf-interpreter\BuildOutput\THUMB2FP\DS55.05\le\FLASH\release\MCBSTM32F400\obj\DeviceCode\Cores\arm\Processors\CortexMx\ItmPort.obj" does not exist.
Task "Message"
[...]

PR #400, opened 2016-02-23
Adds support to use DS5 as Build Tool.

doingnz:
Adds support to use ARM DS5 tool chain. Mostly a simple rework of the existing MDK support. (e.g. scatter files are the same)

This PR is a reworked version of #273 that isolates the changes to DS5 tooling only and relative to the current dev branch. 

The SDK and solutions build without error. Not testing on physical hardware as I don't have either evaluation kit. I could purchase the DISCOVERY board if that is required. I am using the ARM tools (RVDS 4.1 currently) to build a pre MF 4.3 version of .Net MF and that runs fine. Hope to move to MF4.3+ once current product is shipping.

### DIFF
--- a/DeviceCode/Targets/Native/STM32/STM32.settings
+++ b/DeviceCode/Targets/Native/STM32/STM32.settings
@@ -19,7 +19,10 @@
   <PropertyGroup>
     <OEMSystemInfoString>Copyright Oberon microsystems, Inc.</OEMSystemInfoString>
   </PropertyGroup>
-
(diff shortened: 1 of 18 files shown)


smaillet-ms:
:+1:  Though I'd like to see a second from someone with the tools and hardware to verify this works on an actual device.

doingnz:
I have ordered a STM32F407G-DISC1 and STM32F4DIS-BB and will test the build when these arrive.

doingnz:
Replaced with #426 which has been tested on ST Eval Board.

Issue #397, opened 2016-02-23
Tweaks in gcc.targets for GCC 5-2015-q4-major


josesimoes:
I've been doing some tweaking in the gcc.target with the lattest GCC version.
Mostly optimizations and adding extra debug information to debug images.
Should I open a new PR for this?

smaillet-ms:
Can you be more specific on the tweaks? What happens if someone doesn't have that particular release? (e.g. will it just work?, gracefully fall back? or crash? )

josesimoes:
Mostly flags asked to optimize the outputs for "bare metal" systems that I've been collecting when going through the documentation and reading forums.
Considering that newlib nano is already separated and optional all the others are pretty straightforward, I would say. They are not exclusive of this version although the compiler flags can produce different results. In the end I don't think there'll be any catastrophic result.

smaillet-ms:
OK, sounds like something worthy of a PR without further discussion then.

PR #401, opened 2016-02-24
Tweaks with general improvements and extra debug information

josesimoes:
- replace gcc with g++ when compiling C++
- add extra flags when building in debug to output more data and symbols 
- add several flags to improve compilation/link (because we are dealing with "bare metal" systems)
- add extra specs that pair with newlib nano targeting "bare metal" systems
- add debug flavor configurations for final linked file
  (tested with GCC 5-2015-q4-major)

see issue #397


smaillet-ms:
:+1:

PR #404, opened 2016-02-26
Update STM32F4 HAL to the latest ST official HAL

josesimoes:
Although working the existing headers for the STM32F4 are outdated. This is a problem when creating new solutions based in STM32F4 devices other than the '407. "Marrying" with current CMSIS headers and code can be also be difficult.
This PR adds the latest official HAL from ST.
The HAL code files were updated were required (mostly UART, SPI and PWM because there are diferences on what modules are available on each device).
The HAL projects were cleaned up.
The templates were updated.
The platform_selector.h in the solutions was updated to use the new device identifier.
A new file with the HAL configuration was added to the solutions.


smaillet-ms:
rejected. 
Please follow the instructions in the [WIKI for contributions and Git workflow](https://github.com/NETMF/netmf-interpreter/wiki/Contributing). This PR contains changes to bit converter and other changes unrelated to the state HAL updates. 

Each change should be as small as possible and isolated to a clean branch to aid in focusing on the problem at hand. Furthermore third party source must not be added into the tree without first discussing it with the core team as there are legal reviews required before accepting such commits.

josesimoes:
@smaillet-ms 
- I've updated this branch from upstream/dev and rebased it before starting the changes, I can see files marked there as changed which are exactly the same as the ones in the NETMF/dev. I don't know why GitHub is marking them as being different.
- I've squashed into a single commit as requested previously.
- The ST HAL code is publicly available. Furthermore the current stm32f4xx.h, stm32f2xx.h, system_stm32f2xx.h and system_stm32f4xx.h are from that same package, just older. The copyright message is there on top of each file, is kept there and it explicitly mentions that it can be redistributed in "source code format". For all those reasons I presumed that there was OK to add those.
- This update in the HAL is related with the FreeRTOS implementation. As you've instructed "Any changes that apply generally to all of NETMF (e.g. not specific to the Free-RTOS porting) should be done against the dev branch so that it can be applied general". I've considered this to be a general one, not specific to FreeRTOS, thus I've submitted the PR against the dev branch.

As a side note: you'll excuse my enthusiasm and being eager to contribute to this awesome project.

smaillet-ms:
No problem, enthusiasm is welcome, contributions are welcome. We just need to be cautious with respect to third party code. Publicly available doesn't mean it's compatible with the lic. of NETMF. In order to take any new third party code we need to review the lic. terms it is distributed under to ensure that it is compatible with that of NETMF. Last thing we want to do is ruffle feathers for violating lic. terms of someone else's code.

Do not include ANY code you did not write yourself in PRs for this project without first getting an OK from the core team. The way we've been dealing with third party libraries recently is to leave place holders in the tree for where the additional code should go with a readme.md that explains what it is, who owns it and how to get it. (i.e. CMSIS works this way) However, even that still requires a legal review to ensure the licenses are compatible before merging it.

The way the old STMxxx HAL code was dropped into the tree was, frankly, awful and not something we really want to repeat. It wasn't easily re-usable or referenced in new projects, It had some of the STM code modified, and some of the CMSIS code, modified. We need to shift to where we don't modify such code, and instead modify the NETMF assumptions, HAL and PAL to better leverage the large base of existing code as-is. NETMF was initially designed and developed  at a time where such support from silicon vendors didn't exist or was of low grade "sample" quality. Thus it had to define everything from power on reset vector forward. Times have changed and NETMF needs to change with it. Ideally NETMF has no requirements on how to implement startup code before getting to main()

Let's think about this as a first step to getting to that ideal. How do we place/layout the code from ST and CMSIS such that they are easily updated and re-used components for delivering NETMF based systems? Can we or should we use CMSIS-Pack in any way to improve the porting process? What about the new mBed 3 code base and its use of Yota? Or even Microsoft's own NUGET? These are things that have been rattling around my brain for a while now and it's probably time to open the conversation on that stuff.

Git has issues with handling of line endings. For reasons of its own it will normalize all line endings to the '\n' form. It is supposed to deal with that without considering the files as changed, however, in my experience it's not very good at doing so. If the original file had mixed line endings then the checksum of the original changed, and it appears as though the git client thinks the file changed. Also, if a file is converted from raw ASCII into UTF-8 with a Byte-Order-Mark then the whole file is considered changed.

josesimoes:
@smaillet-ms understood your explanations about IP and licencing. 

I'll go and remove the code. As suggested I'll do as with CMSIS and leave a placeholder with instructions on how to get the folder contents.

Regarding the STMxxx HAL, you said it all. 
My suggestion would be to follow the current CMSIS implementation: have a top level folder ST_HAL, or something and drop the code there.

About how to source that:
1) STM32CubeMX 4.13.0 has just added support for generating CMSIS-Pack description file. We could look into that. Although that is tailored to the specific device configuration, which may not be desirable...
2) I'm not familiar with mBed yota so I can't give an input about it.
3) Nuget sounds good and I guess the integration with MSBuild is pretty straightforward. But it would require a previously generated Nuget package with the code that belongs to ST, doesn't it? Wouldn't this fall into the concerns that you've started pointing out on your reply above?
4) The simplest approach would be to just drop the source code there.

Question: what are the plans for NETMF? I'm asking to evaluate if there is a point on keep pouring effort into improving things here. How is Lillum progressing? Is NETMF going into maintenance mode? Is it going to keep moving forward?
As you can see I've been working hard on having a GCC solution supporting networking and SSL. I would rather spend time on adding new stuff and improving what's there, but this is a must have for us here. So we choose working on it and share it.publically and ultimately make it available in the main repository. If someone else is willing to contribute that is more than welcomed!

As for Git, I'm still struggling to get my head around it... I can get my job done with TFS and SVN but Git tends to give me a hard time. Everything that is beyond push and pull always tends to complicate and every now and then ends up on a repository or branch completely messed up... I'm trying, bear with me on this.

smaillet-ms:
I hear you on the git thing (I'm personally not a big fan). If you ruthlessly follow the steps In the wiki you should be fine. The most important point is to keep commits small and focused and to keep a distinct branch for each PR. (That last part is the one that seems to cause the greatest pain, some people I know go with a completely separate clone in a different working directory so they can work on multiple things in parallel without worrying about forgetting to git co <branch> and mixing things up)

Llilum is progressing but it still has a ways to go. However, the problems of how we deal with things like the existing lwip, mBedTLS, CMSIS, mBed 2.0, mBed3.0, STM code bases, etc... are actually pretty consistent between Llilum and NETMF. Ultimately I'd love to be able to get to where a single C/C++ code base library for a given board would serve both NETMF and Llilum. (e.g. If we think of the work to solving all this stuff on NETMF as the basis for a Llilum solution then it isn't going to be throw away work)

For expedience on this to get you and your projects moving the Drop in the folders dotnetmf.proj and other customized code with a readme.md to get the rest will do.

In addition, I want to encourage both thinking about and talking about how we should resolve this stuff for the longer term. It really **should** be possible to get NETMF or Llilum up and running on a new board in under a day (as long as there is existing CMSIS or other similarly supported libraries to run on top of). Obviously we're not there yet.

josesimoes:
@smaillet-ms 

I would love contributing to this. Let's discuss it. Your email isn't public, could you please drop me a line so I can reply back for us to organize it?
I'll start writing some thoughts on this.

I believe that - considering what exists today - it's possible to create a solution in under a day. Providing you'll be using the same processor or a variant of it. For a completely new one that's not possible at all, as you say.

smaillet-ms:
Let's keep the discussion in the public so anyone in the community can chime in.

piwi1263:
Does this 'Not being up-to-date' CMSIS apply to llilum as well for the STM32F4xxx devices ?

smaillet-ms:
I'm not sure I follow that. Can you be more specific? This discussion is generally about NETMF, the references to Llilum was my expression of a desire to migrate NETMF to where it shared a common native code library API with Llilum to help in minimizing the effort to migrate towards Llilum.

piwi1263:
As far as I've understood it is the update to the latest ST HAL for STM32F4 devs, since llilum is avail towards STM32F4 devs I was wondering if the latest is used as well ?

smaillet-ms:
The issue here is the inclusion (and thus the need to update) the STM code in the NETMF tree. While Llilum doesn't require CMSIS any more than NETMF does, it does have a common wrapper for any devices implementing the CMSIS defined APIs, which the STM libraries do. Llilum doesn't include the source to the STM libraries, it just has several reference board configurations that show how to reference those libraries. Thus the original issue that began this discussion is not relevant to Llilum. My goal is to shift from trying to fix the issue short term in NETMF to move to where METMF follows the pattern in Llilum, which was designed based on experience of the issues found here on NETMF.

piwi1263:
Understood, I think. Thanks for clearing.

josesimoes:
@smaillet-ms OK, so let's move the discussion to an open issue, instead of doing it on a closed issue.

Anyway I have a couple of thoughts that I would like to share with you that don't belong in this discussion.

Issue #409, opened 2016-02-26
warning LNK4099


josesimoes:
This warning shows when building a solution (for Crypto.lib and others):

Crypto.lib(modes.obj) : warning LNK4099: PDB 'aes.pdb' was not found with 'Crypto.lib(modes.obj)' or at 'E:\GitHub\netm
f-interpreter\BuildOutput\public\debug\Server\dll\aes.pdb'; linking object as if no debug info [E:\GitHub\netmf-interpr
eter\clr\tools\MetaDataProcessor\dotNetMF.proj]

I guess this could be corrected by adding the PDBs to the setup package.

Another options would be to add a LinkIgnore to project to simply remove the warning as the PDBs aren't really required here. I've submitted PR #410 with this proposed change.

smaillet-ms:
This warning comes from the VC linker and cannot be disabled.

PR #410, opened 2016-02-26
Add LinkIgnore to address missing pdb files in crypto libs

josesimoes:
This removes the warning as reported in issue #409.

### DIFF
--- a/CLR/Tools/MetaDataProcessor/dotNetMF.proj
+++ b/CLR/Tools/MetaDataProcessor/dotNetMF.proj
@@ -13,6 +13,8 @@
     <LinkIgnore>4199</LinkIgnore>
     <!-- warning C4091: 'typedef ': ignored on left of 'NGenHintEnum' when no variable is declared -->
     <ExtraFlags>/wd4091</ExtraFlags>
+    <!--LNK4099: PDB 'filename' was not found with 'object/library' or at 'path'; linking object as if no debug info -->
+    <LinkIgnore>4099</LinkIgnore>    
   </PropertyGroup>
   <Import Project="$(SPOCLIENT)\tools\targets\Microsoft.SPOT.System.Settings" />
   <ItemGroup>


doingnz:
I have tried the proposed change. It has no effect on my clean builds of the SDK. i.e. Warnings are reported. 

Searching on WWW indicates this may be impossible, unless the Linker switch is set to treat all warnings as if they are errors. (/wx). have you tested with a clean rebuild (Delete BuildOutput folder.) 

See: http://www.geoffchappell.com/studies/msvc/link/link/options/ignore.htm

### Unignorable Warnings

### Some warning numbers are apparently too important to ignore:

LNK4099 PDB '%S' was not found with '%S' or at '%S'; linking object as if no debug info

(edit - copied wrong bit of the referenced web page first time)

josesimoes:
@doingnz the proposed change was to remove those warnings when building a solution. I haven's test if it has any effect on the SDK.

doingnz:
@josesmoes I would really like to reach the point where all of these and all the other warnings are removed from all the builds, or knowingly ignored. 

My understanding is the msbuild system will not let us ignore/suppress the warning.

josesimoes:
@doingnz it may sound familiar but... "it works on my machine" :smile:

doingnz:
@josesimoes  works nicely for Solutions ;-). I am using ARM DS5 tools to build. The warnings are  ignored. 

Clearly the behaviour of individual linkers used can be different. 
The ARM DS5 linker will ignore the 4099 warnings as requested (same linker as in MDK).
The Microsoft linker will not ignore these warnings.

Thank you for your suggestion. Less "noise" in the build output is awesome.

smaillet-ms:
The ARM linker doesn't generate those warnings as they are unique to the MS PE/COFF format for Windows. Unfortunately the linker does not allow setting that warning to ignore.

Issue #255, opened 2015-07-27
Is STM32F429I-Discovery supported?


FrederikWJohansen:
Hi I have a STM32F429I-Discovery with a small touch screen, will this board be supported? I know I can download a version from ST (version 4.3.0) but there is no Porting Kit, and i would like to add some interop functions to an advance gfx lib i have for this board (written i pure c++).

cw2:
I cannot speak about the official support, but I am working on STM32F429IDISCOVERY solution for NETMF 4.4. The barebone version (no LCD at the moment) will be available in a few days - initially in my repo.

piwi1263:
HI CW2,

you can use the repo https://github.com/veccsolutions/Vecc.Netduino.Drivers.Ili9341 for a start.

You have to adopt the SPI pins to the SPI5 channel, and it is working on the 429 Disco via SPI. See here for further details https://www.ghielectronics.com/community/forum/topic?id=19593&page=5#msg195861

And yes, I'd be very interested to use 4.4. Tried it myself but failed miserably ... 

Cheers,

Peter.

FrederikWJohansen:
@cw2 awesome, look forward to try it out, if you share it :-)!

ghost:
@cw2 : do you have any news about this port or did I miss it ?

cw2:
I've just updated the repo, the solution is available in [cw2-stm32f429idiscovery](https://github.com/cw2/netmf-interpreter/tree/dev-cw2-stm32f429idiscovery) branch. But, that is version from July 2015, I have not worked on it since then and unfortunately I don't remember its state - whether it at least boots or not. Sorry.

smaillet-ms:
If you have a working version that at least boots with TinyBooter we'd be willing to accept it into the repository so that others can use/contribute to it.

ghost:
@cw2 : thank you very much. I will try to put my hands on it as soon as I can !

cw2:
Well, you are welcome. Although I would be a little bit skeptical about it, due to the reasons mentioned above. I'll too have look at it, hopefully this weekend - if I can get it working, I'll rebase the branch onto the latest dev and make a PR...

In case I will not be able to make it or you beat me to it, feel free to make a PR yourself :)

cw2:
I have checked the STM32F429 solution and it does not work. However, with about three fixes and memory layout settings from STM32F4DISCO I was able to get it boot with TinyBooter.

ghost:
[...]

PR #424, opened 2016-03-10
Added STM32F429IDISCOVERY solution

cw2:
This is an initial version of STM32F429IDISCOVERY solution, as discussed in issue #255.

I have verified build with MDK and [GCC 5.2.1](https://launchpad.net/gcc-arm-embedded/5.0/5-2015-q4-major), all TinyBooter commands via MFDeploy, application deployment and debugging in Visual Studio, serial communication over COM1.

Regarding the peripherals, there are not many enabled at the moment, due to the numerous pin conflicts with on-board features, complex GPIO alternate modes configuration and NETMF limitations (e.g. only one I2C, consequent modules etc.). The plan is to add the LCD in the next version...


cw2:
@Becafuel This includes flash memory fix, the previous version could not erase blocks above 0x08100000 (second MB), causing incremental application deployment failures.

Also, I have rebased it onto the latest upstream/dev (sorry about that).

Issue #429, opened 2016-03-19
InventoryHelper.FindObjectByGuid can't find libraries


AlexandrSurkov:
Hello!

In 2010 I and Igor Kiselev made PKStudio - IDE for exploreing .NetMF porting kit.

Main features that it allows now are:
- View Solution as Tree;
- Edit source files;
- Compile projects from IDE without any console manipulations;
- Jump from compile errors to their possitions in source file editor;
- View Libraries, Library Categories, Features, Processors, Assemblies as Trees;
- View it's properties in Property Grid and in Forms;
- Search Libraries, Library Categories and Features by Name;
- Verify components references and GUIDs;
- View realtionships between components in graphic diagrams;
- Convert MFProjects to fully compiling and linking Keil uVision project

I [move it to GitHub](https://github.com/AlexandrSurkov/PKStudio) and now trying to update it to support .NetMF 4.4 repo.

I found that code into netmf-interpreter\CLR\Tools\PlatformDesigner\ComponentObjectModel\InventoryHelper.cs in line 188 leeds that FindObjectByGuid method can't find any librar:

```
 private object FindObjectByGuid(string guid, IList list)
        {
            string key = guid.ToLower();
            if (m_guidToObjectHash.ContainsKey(key))
            {
                object o = m_guidToObjectHash[key];

                if (!list.Contains(o))
                {
                    Console.WriteLine("Warning the list does not contain a global libary");
                }

                return o;
            }

            if (!(list is List<Library>))
            {
                foreach (object o in list)
                {
                    try
                    {
                        string g = (o.GetType().GetProperty("Guid").GetValue(o, null) as string).ToLower();

                        if (0 == string.Compare(g, key, true))
                        {
                            m_guidToObjectHash[key] = o;
                            return o;
                        }
                    }
                    catch
                    {
                    }
                }
            }

            return null;
        }
```

InventoryHelper.cs was changed [in 4.2 comm brunch](https://netmf.codeplex.com/SourceControl/latest#client_v4_2_comm/CLR/Tools/PlatformDesigner/ComponentObjectModel/InventoryHelper.cs). But changes was not moved to dev.

So can InventoryHelper.cs be changed like in in 4.2 comm brunch?

smaillet-ms:
Feel free to make the change and submit a pull request. 
Fair warning though:
[...]

Issue #405, opened 2016-02-26
Update STM32F4 HAL to include latest official HAL from ST


josesimoes:
Although working the existing headers for the STM32F4 are outdated. This is a problem when creating new solutions based in STM32F4 devices other than the '407. "Marrying" with current CMSIS headers and code can be also be difficult.

I have a branch with the latest official HAL from ST.
The HAL code files were updated were required (mostly UART, SPI and PWM because there are diferences on what modules are available on each device).
The HAL projects were cleaned up.
The templates were updated.
The platform_selector.h in the solutions was updated to use the new device identifier.
A new file with the HAL configuration was added to the solutions.

josesimoes:
@smaillet-ms risking asking the obvious:
Assuming that the approach implemented in Llilum was done learning from what might be wrong/not so good in NETMF, can it be implemented in NETMF too? Or is that too complicated or requires a great deal of changes?

smaillet-ms:
Yes, Llilum was built with full awareness of things like mBed and CMSIS, which didn't exist when NETMF was created. The goal was to minimize the amount of HAL like code we define or standardize around. One of the challenges with NETMF is the rather blurry lines between HAL, PAL, CPU and CLR code layering. With Llilum things are more clear because of the C# vs C/C++ split. So, what I think we need to do with NETMF is migrate in stages:
1. Shift to where we have a standard CMSIS-PAL  
   That is provide a single common NETMF PAL that links against a BSP library for a board that implements the CMSIS-CORE and CMSIS-DRIVER APIs for that device. Thus providing clean isolation between the HAL and PAL and getting NETMF out of defining or controlling the startup sequence, which is a source of a great deal of confusion and frustration when porting to new SOC families.  
2. Unify the NETMF PAL with the Llilum managed interop layer APIs  
   Unifying the top level APIs allows us to have a common native BSP used for both runtimes. This helps to make migration a lot simpler when you are ready to use Llilum. 

[...]

PR #434, opened 2016-04-05
Rework solution configuration and ST CMSIS/HAL integration

josesimoes:
This PR addresses the discussions in #405 and also #429.
A couple of warnings:
- it's a work in progress
- I'm aware that it's huge but couldn't find a better way to deal with all the changes because it touches a lot of stuff
- I know it's not perfect, but it's a start 

Having said that: it's a proposal open for review, discussion and criticism.

What's included (and in the works)
1. Add PowerShell script to download and "install" ST CMSIS and HAL files (this addresses the question of not having third party code in the repository).
2. The script is called when building a solution to make sure everything is available (this addresses the question of dealing with automated build processes).
3. Reworked the way the hardware features are added to a solution to make developer's life easier and not requiring a deep knowledge of what is "under the hood".
   1. The device series/variant is moved from platform selector to solutions settings.
   2. TinyBooter and TinyCLR each have a settings file to allow granular selection of what features are to be added on each flavor. These settings files merge the solution settings and each is the one that goes into the respective proj file (instead of the global one as before).
   3. These settings files have a simple True/False selector to choose what should be added. These properties latter reflect what is compiled/linked in the final image. Having this selection here and as a property allows them to be visible and accessible in the projs, build system and code.
   4. The features property allow that (using #defines) the calls to the HAL init/deinit functions are added or not according to the feature selection. 
   5. There is a readme file with the instructions on how to build a new solution.
   6. The  TinyBooter and TinyCLR projs are organized in a way that they don't need to be (almost...) touched when building a new solution. Unless a very particular change is required, of course.
   7. There is a HAL project (one for CLR and another for loaders) that includes all the features.
   8. All the stubs are now implemented as 'weak' functions. This makes things easier because it allows that all the stubs can now be compiled and always added by default because the linker will replace  the 'weak' ones with the 'strong' implementation (when it exists) and the optimizations will remove the 'dead' code when they are not used.
   9. All this makes it very easy to add new series and MCUs (at least for ST parts, right now there is only the HAL for the F4xx series but any other series can be added with minimal effort by using the ST Cube files and taking advantage on the new organization).
4. The "new" HAL is total based in the CMSIS and HAL code supplied by ST.
5. The individual projects for the HAL features are now based in the ST HAL and follow the ST guidelines and philosophy.
6. The hardware configuration is now simpler because it uses the HAL naming and defines. This makes it easier to add new series/variants becuase the code knows what is available and what isn't and it adapts accordingly. E.g how many and which USARTS, SPI port, which pins belong where, etc.
   (GPIO, USART, SPI, I2C, power, flash and time are implemented; DA, ADC, PWM and USB are on the works)
7. CRC calculations are selectable between the software implementation and hardware based.   
8. Most of the configurations and HAL options are now in the system_stm32f4xx.c in the solution folder, replacing what was in platform selector.
9. USB support (using ST middleware libs) should be much easier for both device and host. We are close to have USB connection for MFDeploy. It currently enumerates the NETMF device and receives the debug packets. The last part (replying to MFDeploy) is in the works.  
10. The scatterfiles were reworked to use of the newlib nano and implement the stack/heap. The managed heap size is now automatic.
11. The "simple heap" was removed because the modern compilers can provide that and we can free NETMF from doing it. (There is still a small remapping required for the crypto libs).

Final notes:
- The changes are being included in both MCBSTM32F400 and STM32F4DISCOVERY. We only have available the STM32F4DISCOVERY board so we can't test the other hardware.
- Right now only the TinyBooter is compiling and tested (except USB).


smaillet-ms:
You'll need to split this up into smaller PRs before we can consider it, there are 113 commits across 200+ files and it brings the GitHub site to it's knees when attempting to review or provide feedback.

smaillet-ms:
Also FYI, there are already powershell scripts for downloading the CMSIS and binary tools. And I'm working on another to download and install the crypto libs (Though that can't actually be done for an automated build service running in a restricted account it will allow building a single powershell script to get all the pre-requisites for individual developers)

josesimoes:
@smaillet-ms  I can go and try to squash that into a single commit... does it help? 
Because this is a really big change touching a lot of files. Not sure how small/atomic I can make those smaller PRs for each individual one to make sense...

Don't know if you are familiar with it but there is an awesome tool for Git: [Git Extensions](https://github.com/gitextensions/gitextensions). It works great for this kind of work that tends to be impossible to manage on the web interface. You may want to give it a try. There is a tool for working with PRs (no matter how big they are!) :wink: 

Regarding the PowerShell script: I'm aware of it The thing is that that script downloads the CMSIS "general" distribution. The one that is there on the PR is different: it downloads the CMSIS implementation for STM32 devices and the ST HAL files. 

As for running the ps scripts in a build service I believe it can be accomplished by signing the scripts. I could have signed the one in the PR but it would have our company signature. I believe it should be possible to include the ps scripts in the NETMF sign process and have them signed with Microsoft certificates.

smaillet-ms:
PS scripts in the build service aren't the issue - that's actually built into the VSO builds we are now using, it's the fact that we are required to distribute the crypto libs as an MSI that's the problem. When a build service is set up it doesn't (and frankly shouldn't be) run with permissions to install software on the server, so it can't run an MSI. 

So you are also pulling the ST packs etc... Can we hold on this then?

We already have a lengthy discussion on the build system and how we want to solve the problem of these types of dependencies. I've published the start of the [CMSIS-Pack .NET library](https://github.com/NETMF/CMSIS.Pack) for managing a pack repository from C# so we can have a new app that is a cross between the pack manager in uVision and the old PKStudio and build that into VS. I'd rather not do that kind of multiple times.

josesimoes:
1) This is maybe sound like a dumb question, but... can't you just distribute the crypto libs with the plain files? Without the MSI and requiring an installation? Or even better, as a Nuget package? For me this would address this automated build/install/etc hurdle.

2) Yes that PS script is able to download and install any ST pack for any of the available series.

3) I saw your new CMSIS package manager last Friday but hadn't a chance to look into it yet. From your description above it will be usable from VS at some point, correct? For that to be usable wouldn't it require a way to manage the building a .NETMF Solution as a VS solution?

josesimoes:
@smaillet-ms I would be very interested on your comments on the way we are proposing to manage the features availability in a Solution. I really believe it eases (a lot!) regarding what a developer needs to touch and understand about Solutions architecture and not having to touch the prj files and all that.

smaillet-ms:
I'll be looking into that on this PR. Please note I'm not in any way trying to suggest there's nothing of value here. There's a lot of different angles and aspects to be considered for a robust architecture neutral system. We've got a system now that was the result of multiple whiplash speed design/implementations and I'd like to avoid repeating any mistakes. Furthermore we've got some bugs that are blocking real world use in SSL and some VB issues to knock out (likely requiring a 4.4.1 release at least). I'd like to hit the bigger build stuff right after those are done.

josesimoes:
Understood. 
Know that we are doing this for real world implementations too. Our company develops commercial products based in .NETMF.
We are here because we wanted to contribute back to the community for all the great stuff that we've been receiving so far. And because we love .NETMF and believe it's a truly remarkable platform.
All this started because we needed networked SSL connectivity and we are using GCC.
First approach was to use FreeRTOS instead of RTX, which we did.
Next we've stumbled on the need to update the HAL to the new ST stuff and then the well known issues with networked connectivity and SSL.
In this PR we've gone through most of the HAL stuff. Right now the USB connection to MFDeploy needs a final debug and it will be usable. Next week we'll start adding the network stuff including the LwIP and SSL.
All this to say that we are activelly working on this and we are under a tight schedule for a project we need to deliver. If we are able to mix and match both our teams efforts: excelent. If not, so be it but we really need to keep moving forward with this. At a latter time some of our work maybe usefull.

smaillet-ms:
Understood. I hate duplication and re-working something multiple times but sometimes the real world gives you little choice. Having something with some field time on it isn't a bad thing either. It will give you a chance to evaluate the choices you made and see how well they hold up in the real world rather than just on paper. That will help drive to a better overall long term experience.

It should be a lot easier to put NETMF onto existing Cortex-M devices given how much is available in CMSIS now. And for any other processor type there are similar types of libraries available now to work from as well.

Issue #376, opened 2016-01-28
Device not showing in MFDeploy


DerekGn:
I have implemented a USB driver for the SAMA5D3 xplained board. The device enumerates correctly and shows up in device manager as a WinUsbDevice.

Below is the device information:

```
Information for device SAMA5D3XPLN (VID=0x03EB PID=0xA08F):

Connection Information:
------------------------------
Connection status: Device connected
Device actual bus speed: HighSpeed
Device is hub: No
Device address: 0x0002
Current configuration value: 0x01
Number of open pipes: 2

Device Descriptor:
------------------------------
0x12    bLength
0x01    bDescriptorType
0x0200  bcdUSB
0x00    bDeviceClass   
0x00    bDeviceSubClass   
0x00    bDeviceProtocol   
0x08    bMaxPacketSize0   (8 Bytes)
0x03EB  idVendor
0xA08F  idProduct
0x0200  bcdDevice
0x01    iManufacturer   "Microsoft OpenTech "
0x02    iProduct   "SAMA5D3XPLN"
0x00    iSerialNumber
0x01    bNumConfigurations

Device Qualifier Descriptor is not available. Error code: 0x00000001

Configuration Descriptor:
------------------------------
0x09    bLength
0x02    bDescriptorType
0x0020  wTotalLength   (32 Bytes)
0x01    bNumInterfaces
0x01    bConfigurationValue
0x00    iConfiguration
0xC0    bmAttributes   (Self-powered Device)
0x32    bMaxPower   (100 mA)

Interface Descriptor:
------------------------------
0x09    bLength
0x04    bDescriptorType
0x00    bInterfaceNumber
0x00    bAlternateSetting
0x02    bNumEndPoints
0xFF    bInterfaceClass   (Vendor specific)
0x01    bInterfaceSubClass   
0x01    bInterfaceProtocol   
0x00    iInterface

Endpoint Descriptor:
------------------------------
0x07    bLength
0x05    bDescriptorType
0x81    bEndpointAddress   (IN Endpoint)
0x02    bmAttributes    (Transfer: Bulk / Synch: None / Usage: Data)
0x0040  wMaxPacketSize   (64 Bytes) 
0x00    bInterval

Endpoint Descriptor:
------------------------------
0x07    bLength
0x05    bDescriptorType
0x02    bEndpointAddress   (OUT Endpoint)
0x02    bmAttributes    (Transfer: Bulk / Synch: None / Usage: Data)
[...]

PR #439, opened 2016-04-20
Correct code with wrong call to USB_OpenStream

josesimoes:
If USB_ALLOW_CONFIGURATION_OVERRIDE is defined and the call to USB_Initialize has a debugger port in the argument, USB_OpenStream is already called there (@ line 265). After a successful return from Initialize the call to USB_OpenStream will fail because the stream is already opened.

Note: I've stumbled onto this when debugging the CMSIS USB driver I'm working on. The result is that the USB device is enumerated and shows in the device manager but it's not picked up by MFDeploy or VS because after the initial enumeration the USB module is too busy being initialized again and again. This may explain the behaviour reported in issue #376.
The root cause is that the call sequence (...) DebuggerPort_Initialize -> USB_Initialize is repeated over and over from CLR_HW_Hardware::ProcessActivity() because the debugger port never has the chance to initialize. This is not 100% repetible. It seems to occur only if the USB init falis for some reason when sending the initial debug ping message, which call the above sequence. After that the  ProcessActivity() starts calling that sequence again and most likely a race condition occurs with the USB stream.

### DIFF
--- a/DeviceCode/pal/COM/ComDirector.cpp
+++ b/DeviceCode/pal/COM/ComDirector.cpp
@@ -16,10 +16,17 @@ BOOL DebuggerPort_Initialize( COM_HANDLE ComPortNum )
             if(USB_CONFIG_ERR_OK != USB_Configure( ConvertCOM_UsbController(ComPortNum), NULL ))
                 return FALSE;
 
-            if(!USB_Initialize( ConvertCOM_UsbController(ComPortNum) ))
-                return FALSE;
-
-            return USB_OpenStream( ConvertCOM_UsbStream(ComPortNum), USB_DEBUG_EP_WRITE, USB_DEBUG_EP_READ );
+            if(USB_Initialize(ConvertCOM_UsbController(ComPortNum)))
+            {
+#if defined(USB_ALLOW_CONFIGURATION_OVERRIDE)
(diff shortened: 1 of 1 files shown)


smaillet-ms:
USB_ALLOW_CONFIGURATION_OVERRIDE is technically orthogonal to the use of USB as a debug transport. Your PR here is merging the two concepts, which is something we don't want to do. USB_ALLOW_CONFIGURATION_OVERRIDE is a build time configuration to allow MFDeploy and managed apps to change the USB configuration from what it booted as. While this is obviously something you don't want while using USB AS your debug transport it isn't unique to that scenario. 
A cleaner, long term solution would be to make the initialization process consistent so we don't have the multiple initialization calls, or add a requirement that the initialization is [idempotent](https://en.wikipedia.org/wiki/Idempotence)

josesimoes:
WOW 😲 

There is clearly an issue with the call to USB_OpenStream after a successful return from USB_Initialize.
That call can - and will - fail depending on how fast things are moving.
The outcome is that it will lead to an unresponsive USB device because it will enter into a loop of initializing it over and over.
The bug is identified so feel free to implement a correction that you see appropriate.

smaillet-ms:
Closing this PR I've opened bug #476 to cover the issue.

Issue #451, opened 2016-05-13
Wiki article on how to run Unit test please


doingnz:
It is necessary to make a debug build of the SDK to run the tests?

I can checkout netmf, run build_sdk and build_solution for STM32F4DISCOVERY using the default build environment. TinyBooter and TinyCLR load into the STM32F4DISCOVERY.

When I then try execute RunTests.exe with no parameters, it throws up errors. 
Same errors if I specify command line parameters to direct the test to be via USB to the device. 

```
ERROR: System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.SPOT.Platform.Test.TestSystem.RunMFTests.get_InstallRoot() in C:\Projects\BPplus\netmf-dev\netmf-interpreter\Test\Pl
atform\Tools\MFTestSystem\TestSystem.cs:line 957
   at Microsoft.SPOT.Platform.Test.TestSystem.RunMFTests.SetResultPaths() in C:\Projects\BPplus\netmf-dev\netmf-interpreter\Test\Pla
tform\Tools\MFTestSystem\TestSystem.cs:line 1316
   at Microsoft.SPOT.Platform.Test.TestSystem.RunMFTests.Run(String[] args) in C:\Projects\BPplus\netmf-dev\netmf-interpreter\Test\P
latform\Tools\MFTestSystem\TestSystem.cs:line 562
   at Microsoft.SPOT.Platform.Test.TestSystem.RunTests() in C:\Projects\BPplus\netmf-dev\netmf-interpreter\Test\Platform\Tools\MFTes
tSystem\TestSystem.cs:line 125
TestSystem: An uncaught exception was thrown when invoking RunTests: System.NullReferenceException: Object reference not set to an i
nstance of an object.
   at Microsoft.SPOT.Platform.Test.TestSystem.RunMFTests.get_InstallRoot() in C:\Projects\BPplus\netmf-dev\netmf-interpreter\Test\Pl
atform\Tools\MFTestSystem\TestSystem.cs:line 957
   at Microsoft.SPOT.Platform.Test.TestSystem.RunMFTests.SetResultPaths() in C:\Projects\BPplus\netmf-dev\netmf-interpreter\Test\Pla
tform\Tools\MFTestSystem\TestSystem.cs:line 1316
   at Microsoft.SPOT.Platform.Test.TestSystem.RunMFTests.Run(String[] args) in C:\Projects\BPplus\netmf-dev\netmf-interpreter\Test\P
[...]

PR #452, opened 2016-05-15
Test SDKs

smaillet-ms:
- Fixed Test Framework MSI generation so it builds for the correct version
- Added script and test list to run mscorlib tests as run on daily builds for verification of the core runtime engine

Related to: #451


doingnz:
Thank you for putting in time over the weekend on this! greatly appreciated. 

A suggestion is to move the file(s) with the list tests to the Test folder. Less to clutter in the root of the repository as there are likely to be additional lists of test for graphics, file system, etc..  

Also consider adding an argument check to dotest.cmd and example usage (See suggestion below). Another approach is to default to the mscorelibtests.txt if it exists when no argument is supplied. 

```
@echo off
if NOT "%1"=="" GOTO :ARGSOK

:ERROR
@echo.
@echo Error: Invalid Arguments!
@echo.
@echo Usage: dotest TEST_LIST
@echo.
@echo where  TEST_LIST  is a text file containg a list of paths to .sln files of the tests to execute.
@echo                   Single path to .sln file per line.
@echo. 
@echo                   Example contents for the file specified by TEST_LIST
@echo                           Test\Platform\Tests\CLR\mscorlib\arrays\arrays.sln
@echo                           Test\Platform\Tests\CLR\mscorlib\attributes\attributes.sln
@echo                           Test\Platform\Tests\CLR\mscorlib\basicconcepts\basicconcepts.sln
@echo.
@echo     e.g.  dotest.cmd mscorelibtests.txt
@echo.
GOTO :EOF

:ARGSOK
setlocal
call setenv_vs.cmd 14
@echo on
BuildOutput\public\Release\Test\Server\dll\RunTests.exe -testlist %1 -transport Emulator -device Microsoft
endlocal
```

smaillet-ms:
The test scripts are AS-IS from our internal build infrastructure. The goal is to produce a daily build that is not based on any internal proprietary tooling that is easily replicated/mirrored by others. Once that is functional we'll look to better optimize/improve. For now I want to limit the differences as much as possible to the infrastructure support itself. So it is easier to compare and contrast when results are different.

PR #458, opened 2016-05-22
qualify path to CreateCLRDefines.proj

doingnz:
ensure the correct path is used to reference the project file by adding `$(SPOCLIENT)` to the path defined in the Directory Property.


smaillet-ms:
What bug is this supposed to be fixing? In the future please provide a reference to a bug in pull requests. If no bug exists then one should be created BEFORE submitting the PR as the issue may not be what it seems on the surface.

The <Directory> property in the project files is intended to be relative - in what conditions or case are you seeing a problem that this needs to be changed?

PR #465, opened 2016-06-08
Fix GPIO_Buttons_Compile_Error

techcap:


smaillet-ms:
Please do not post a pull request without having an issue filed that it is supposed to fix.
It is unclear from this PR what the bug is that is intended to be fixed the code built fine previously without errors so the commit comment about fixing a compiler error is suspect. Furthermore this is altering the code path and computations of time in no small way so, again is suspect. Let's discuss the actual problem and be sure one exists before taking this.

techcap:
There is no Time_TicksToTime method. if I use timewarp option, I get compile error. So I changed to HAL_Time_TicksToTime.

smaillet-ms:
I would say - Don't use the timewarp option - it is an ancient leftover from an old emulator model that is no longer relevant/supported (some parts of the code that have had work since then have removed any reference to it but others still remain unaltered as they haven't had any other work needed).

techcap:
I found the option during testing my new device because of sleep problem. And finally, I solved my h/w problem and I don't need that option any more. Thank you very much^^

Issue #474, opened 2016-07-06
Successfully build Discovery4 with GCC 5.4.1


josesimoes:
GNU ARM Embedded 5-2016-q2-update with GCC 5.4.1 was released last week.
After a few tweaks I've successfully build a NETMF image.

Image size comparison
- GCC 4.9
  Tinybooter.bin 41812B
  ER_FLASH  322525B
- GCC 5.3.2 with newlib nano and default compile settings
  Tinybooter.bin 42700B
  ER_FLASH  308380B
- GCC 5.3.2 with newlib nano and C11 compile settings
  Tinybooter.bin 42220B
  ER_FLASH  305276B

smaillet-ms:
Not bad. 😄 
TinyBooter is oddly a bit larger in 5.3.2 with newlib nano... That's unexpected, any idea what is making it bigger?

~~Is that C11 or C++11? - There's a difference, especially since most of NETMF is actually C++ code (though most of it doesn't really use much of the C++ language at this point)~~ Never mind, PR #475 shows both are set (Although it applies globally to all projects and solutions so that needs to be made opt-in before it can be merged into the repo)

piwi1263:
I just can confirm that after applying the changes from José a release build for Tinybooter and the TinyClr deployed succesfully to the STM32F429IDISCOVERY as well. 
- GCC 5.4.1
- Tinybooter.bin -> 42572B
- ER_FLASH -> 303868B

cw2:
@smaillet-ms The extra code in 5.4.1 comes from newlib-nano malloc locking, CRT _init and _fini functions and minor differences in generated assembly instructions and alignments. However, the majority is endian swapping routines, which are not present in 4.9 build, namely (symbols from Tinybooter.dump):

```
_Z10SwapEndianj
_ZN10WP_Message10SwapEndianEv
_ZN13Loader_Engine10SwapEndianEP10WP_MessagePvib
_ZN13Loader_Engine17SwapEndianPatternERPhjj
```

which is very weird, because they are conditionally compiled #ifdef BIG_ENDIAN and should not even be there. This seems to be a problem of GCC 5.4.1, because those functions are not present in any of the older builds (4.9.3, 5.2.1, 5.3.1).

smaillet-ms:
[...]

PR #475, opened 2016-07-06
Fix to build Discovery 4 using GCC 5.4.1

josesimoes:
- add new symbol to missing symbols for gcc scatter file
- add new-lib nano setting to tinybooter proj
- add platform emulated setting to solution settings
- add compiler flags to enable C11


Solutions/STM32F4DISCOVERY/TinyBooter/TinyBooter.proj
```diff
@@ -23,6 +23,8 @@
         <ExtraTargets>$(ExtraTargets);CompressBin</ExtraTargets>
         <ScatterFileDefinition>scatterfile_bootloader_$(COMPILER_TOOL).$(SCATTER_EXT)</ScatterFileDefinition>
         <EXEScatterFileDefinition>scatterfile_bootloader_$(COMPILER_TOOL).$(SCATTER_EXT)</EXEScatterFileDefinition>
+        <!-- newlib-nano does not support long-long and long-double formatting, use standard newlib instead -->
```
smaillet-ms, line 26: What is it in Tinybooter that is requiring long-long and long-double formatting? Perhaps we should fix/remove that so we don't need to do this and can use newlib-nano?

Well actually scratch that, it seems like the comment is redundant as it is setting NewLibNano == true for GCC...

tools/Targets/Microsoft.Spot.system.gcc.targets
```diff
@@ -52,7 +52,7 @@
     <CC_CPP_TARGETTYPE_FLAGS Condition="'$(FLAVOR)'=='Instrumented'" >$(CC_CPP_TARGETTYPE_FLAGS) -O0</CC_CPP_TARGETTYPE_FLAGS>
     <CC_CPP_TARGETTYPE_FLAGS Condition="'$(FLAVOR)'=='Release'"      >$(CC_CPP_TARGETTYPE_FLAGS) -Os</CC_CPP_TARGETTYPE_FLAGS>
 
-    <CC_FLAGS>$(CC_FLAGS)-xc</CC_FLAGS>
+    <CC_FLAGS>$(CC_FLAGS)-std=c11 -xc</CC_FLAGS>
```
smaillet-ms, line 55: This is a global change impacting ALL solutions building with GCC, thus needs some special care so we don't break working builds arbitrarily. The settings added should be conditional (e.g. opt-in from the project)

Solutions/STM32F4DISCOVERY/TinyBooter/TinyBooter.proj
cw2, line 26: IIRC, the Tinybooter code does not require long-long and long-double formatting support, but the TinyCLR does (uses "%I64" and "%lld" printf formats). The newlib-nano included in GCC toolchain does not have those options enabled (i.e. long.ToString() and double.ToString() does not work correctly), so there are basically two options:
1. Use default newlib-nano in TinyBooter and standard newlib in TinyCLR,
2. Build a custom newlib-nano with long-long and long-double formatting support and use it in both TinyBooter and TinyCLR.

The ideal solution would be to use default newlib-nano in Tinybooter and a custom newlib-nano in TinyCLR - I guess that could be done with some additional tweaking of linker settings, but it would probably be too confusing...

Unfortunately, I have not yet had a chance to review the recent GCC versions, sometimes they introduce changes in newlib-nano, so if they for example added long-long and long-double formatting by default, it could made things a little bit easier.

smaillet-ms, line 26: Another option would be to change the CLR code so it doesn't use printf style formatting from the CRT for long and double and instead does it's own thing. Thus, allowing for newlib-nano in all cases. It would be ideal if we could figure out how to eliminate the duplication of the CRT completely (e.g. Tinybooter has a copy of the CRT and so does TinyCLR ) do we really NEED two copies of the CRT in the device?)

cw2, line 26: The obvious/naïve solution to eliminate Tinybooter's CRT is to eliminate the Tinybooter itself :)

smaillet-ms, line 26: A fair point, though it presumes that all devices have some sort of firmware update support built into the silicon to allow updates even if the main application image is completely hosed.

Doing that for NETMF requires some major re-work of the wireProtocol and debugger communication layers (and MFDeploy). Effectively abstracting out the entire wire protocol so that a plug-in of some sort could manage all the actions needed by MFDeploy to update the NETMF image. Unfortunately most such facilities built into silicon and devices requires some sort of external hardware/jumper to engage the update. (Although for development purposes the abstraction could also allow for standard JTAG/SWD/CMSIS-DAP access too which would provide full access and control [Yeah, I've been thinking about this one a while!])

Solutions/STM32F4DISCOVERY/STM32F4DISCOVERY.settings
```diff
@@ -16,6 +16,7 @@
     <IsSolutionWizardVisible>True</IsSolutionWizardVisible>
     <ENDIANNESS>le</ENDIANNESS>
     <NO_BOOTLOADER_COMPRESSION>true</NO_BOOTLOADER_COMPRESSION>
+    <PLATFORM_EMULATED_FLOATINGPOINT Condition="'$(COMPILER_TOOL)'=='GCC'">true</PLATFORM_EMULATED_FLOATINGPOINT>
```
cw2, line 19: I understand this is required to enable use of newlib-nano, but is it really worth it to use emulation on a CPU with a hardware FPU just to save a few percents of the firmware size? 

ER_FLASH size built with GCC 5.4.1 and standard newlib is 327084, which is about 6% larger than 308380 reported in issue #474; the practical difference is even smaller due to the block size in the flash memory code area.

How about moving this setting to a separate 'size-optimized' solution?

Solutions/STM32F4DISCOVERY/scatterfile_gcc_missing_symbols.xml
```diff
@@ -1,5 +1,10 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <ScatterFile xmlns="http://schemas.microsoft.com/netmf/ScatterfileSchema.xsd">
+  <!-- 
+    - using newlib-nano adds libnosys.a, which calls sbrk, which requires the 'end' symbol
+    - this is pulled in by a _printf imported in function _Crypto_StepRSAOperation @ Crypto.lib(rsa_zenith.obj) 
+  -->
+  <Provide Name="end" />
```
cw2, line 7: Shouldn't be the 'end' symbol added to the linker script (scatterfile) instead?

It needs to be placed at the lowest allocatable memory address not occupied by the program or its data.

cw2, line 7: On the second thought, why is crypto.lib being linked and not stubs? The linker error appears also with GCC 4.9, which IIRC worked just fine back then. I don't remember ever installing/using crypto libraries, perhaps something else has changed that results in this (missing 'end' symbol) error (? )

Solutions/STM32F4DISCOVERY/STM32F4DISCOVERY.settings
josesimoes, line 19: Actually this because of the crypto libs. GCC link complains about a mismatch in the VFP flags between the crypto libs and the netmf. This has been discussed a number of times on other threads.

Solutions/STM32F4DISCOVERY/TinyBooter/TinyBooter.proj
josesimoes, line 26: I vote for changing the ToString implementation instead of sacrificing the newlib nano use.

Solutions/STM32F4DISCOVERY/scatterfile_gcc_missing_symbols.xml
josesimoes, line 7: Correct me if I'm wrong but I've traced the end symbol requirement caused by a (possibly) forgotten printf in the crypto libs. See the comment above the end line above. This takes us back to the crypto libs, again. I really think that should be addressed and several pending issues will be gone.

tools/Targets/Microsoft.Spot.system.gcc.targets
josesimoes, line 55: My understanding is that switching to c11 adds only improvements in the compiling and linking. As the code base is "old" there are no new stuff brought with c11 that could break things. If that happens those issues can be corrected individually. Moving to c11 definitely improves the code size.

Solutions/STM32F4DISCOVERY/scatterfile_gcc_missing_symbols.xml
cw2, line 7: Thanks for the explanation.

I am a little bit surprised that the printf is referenced from crypto function, because if I understand it correctly, the solution should not use it (I definitely don't have crypto.lib anywhere, and the one in the build output is copy of crypto_stub.lib; crypto dotNetMF.proj is somewhat weird).

I'll have a closer look at it, maybe I am wrong...

Nonetheless, the proprietary crypto code should be replaced with an appropriate alternative (along with OpenSSL). Perhaps it's time to check out mbed TLS...

Solutions/STM32F4DISCOVERY/STM32F4DISCOVERY.settings
cw2, line 19: Ah, I have found the issues, thanks.

So, if I understand it correctly, this happens only when crypto.lib is used (resp. installed)? Because IMHO it would be much better to fix the cause in crypto.lib (as you suggest in the other comment) than modifying floating point setting (which would be needed for all solutions configured to use newlib-nano).

Solutions/STM32F4DISCOVERY/scatterfile_gcc_missing_symbols.xml
josesimoes, line 7: That seems to be the cause. If you build the tinybooter with GCC you'll see the error with the description going down all the way to a printf in a lib from crypto.
I'm 100% with you that this crypto stuff should be replaced for good.

smaillet-ms, line 7: A little while ago I had a look at mBedTLS and it seems to offer up almost everything we need on the device side. On the desktop side the standard .NET framework libraries have matching counterparts. Unfortunately there's one snag. While I've been able to get a mBedTLS and .NET Framework to correctly match sign and verify operations with the RSA keys the hash algorithm the cryptolibs use in creating a signature doesn't appear to be a standard one. I've written some test utils to use the cryptolibs to compute the hash used with the RSA keys for creating the signature and have not found any standard hashing algorithm that matches the output of the cryptolibs. 

Thus, while it is possible to completely replace them (and OpenSSL) with mBedTLS on the device and normal .NET framework on the desktop side, there's currently no way to do so in a backwards compatible manner. Thus we'd need to drop Back Compat support or drag around the old desktop builds of the cryptolibs for use by MFDeploy when dealing with older devices.

cw2, line 7: That's rather great news (that you've already got mbed TLS working with NETMF) - can you merge it? :innocent: 

The question is, how many devices actually use the proprietary crypto hash function?

I guess leaving crypto libs for another minor version release is not much of an issue, they can be always removed later.

Edit: Perhaps the whole contents of \crypto folder can be moved to .NET Micro Framework Cryptographic libraries installer package.

smaillet-ms, line 7: Ha, no it's not done work. Just an investigation to see about using mbedTLS to create an OSS implementation of the crypto lib support. 

As to how many devices use the crypto support I don't know. If they use the firmware image signing and TinyBooter that requires the cryptolibs. I believe, but haven't yet verified that the MFUpdate mechanism doesn't use the crypto libs signature/hashing so could be replaced. At this point I'm leaning towards just pulling the old crypto support as there's a lot in the mix for the next round of NETMF to warrant a new major label (for example v5.0.0 ) where a lot of legacy cruft can be ejected. But much of that depends on actual users to provide feedback on what is used/needed.

Issue #473, opened 2016-07-04
Remove "custom heap"


josesimoes:
Considering that:
- modern compilers/runtimes are more than capable of handling memory management
- the use of newlibnano require that an implementation of sbrk and other system calls is provided (see https://sourceware.org/newlib/libc.html#Syscalls)

I suggest that the low level memory management (custom heap) should be removed from NETMF and replace it with the "standard" memory management support.

PS: this has already been addressed in #417 not to mention that it looks pretty obvious considering what exists today.
So here it is the issue before the PR.

smaillet-ms:
👍

josesimoes:
Is this a yes? ☺ 

This, along with other improvements, would address the end symbol being discussed oh the that other PR.

smaillet-ms:
Yes, the thumbs up is an approval. I've always hated that NETMF did it's own thing instead of using a standard compilers run time library. Unfortunately when NETMF was first created those libraries were quite bloated and not well tuned for embedded micro controller use. Now with MDK uLib and newlib-nano it's a very different world.

PR #484, opened 2016-08-04
Fixes issues with endian macros (in GCC)

### LINKED ISSUE #481: wieredness with BIG_ENDIAN macro
Starting with GCC 5.4.1 the BIG_ENDIAN define seems to be defined internally in the compiler.
This has been reported by @cw2 in #474. I can confirm that I'm seeing this too on my builds.

I've opened a ticket with GNU ARM Embedded Toolchain. Waiting on their comments.
Thinking on a solution for this, I found that CMSIS has a macro defined for this purpose (see "Pre-processor Macros" in [http://www.keil.com/pack/doc/cmsis/DSP/html/index.html](url). It's ARM_MATH_BIG_ENDIAN. As we seem to be moving towards CMSIS and we are already using other macros from there (__FPU_PRESENT for example), suggest that the current BIG_ENDIAN in NETMF is replaced with ARM_MATH_BIG_ENDIAN.

I can submit a PR for this immediately.

josesimoes:
Fixes #481


josesimoes:
Can this be merged now?
I'm working on #473 and have to have this workaround to be able to build and test on hardware.
(I already have it there - otherwise it wouldn't be testable - but was trying to avoid mixing this in that branch so it won't get into the PR)

smaillet-ms:
Done.

PR #485, opened 2016-08-05
Reverted change from previous commit that broke GCC builds

cw2:
Removed nosys.specs that caused "undefined reference to 'end'" error in _sbrk function.

This could be considered a minimal-changes hotfix (affects all GCC builds of all solutions and toolchain versions, tested 4.9 to 5.4), until another solution is implemented (e.g. as proposed by @josesimoes in his PRs).

### DIFF
--- a/tools/Targets/Microsoft.Spot.system.gcc.targets
+++ b/tools/Targets/Microsoft.Spot.system.gcc.targets
@@ -221,7 +221,7 @@
     <GCCSTDLIBS>-lstdc++ -lsupc++ -lm -lgcc -lc </GCCSTDLIBS>
 
     <!-- Use newlib-nano (default) or newlib -->
-    <LINK_NANO_SPECS Condition="'$(NewlibNano)' != 'false'">-specs="$(GNU_LIBGCC_DIR)\nano.specs" -specs="$(GNU_LIBGCC_DIR)\nosys.specs"  </LINK_NANO_SPECS>
+    <LINK_NANO_SPECS Condition="'$(NewlibNano)' != 'false'">-specs="$(GNU_LIBGCC_DIR)\nano.specs" </LINK_NANO_SPECS>
   </PropertyGroup>
 
   <!-- item group  -->


smaillet-ms:
Do we have a detailed explanation on what is going on here? The nosys.spec is supposed to be designed for newlib and bare metal systems (e.g. NO SYStem) so disabling that seems counter-intuitive and suggests a deeper problem we don't understand.

cw2:
Ok.

Adding nosys.spec also requires providing the `end` symbol (in the linker script, i.e. the scatterfile). Unfortunately, this has not been done, so the result is broken GCC build for all solutions.

Removing nosys.spec fixes this problem for all solutions and their GCC builds. Alternatively, every GCC scatterfile would have to be modified to provide the `end` symbol. Since I cannot verify all solutions due to the lack of the hardware, I decided to go with reverting/removing nosys.spec (which was the state before the change that broke the GCC builds).

smaillet-ms:
Is this 'end symbol' something new to the latest GCC? We definitely built and ran tests on MCBSTM32F400 using GCC so it was working there at one point.

cw2:
No, the missing 'end' symbol is not related to any particular GCC version, it is caused by adding libnosys.

The GCC builds were working until commit https://github.com/NETMF/netmf-interpreter/commit/b5adf76540a2213ea31338e728cfef2b345e2195#diff-149cd7cd15f32c47c7d3900f4bd3350f added nosys.specs that broke them. At the moment, no solution can be built successfully with any GCC, checked with 4.9 to 5.4.

To summarize,
- GCC builds were working,
- Adding nosys.specs resulted in error "undefined reference to `end'",
- To fix that either remove nosys.specs or provide the 'end' symbol in linker scripts (scatterfile).

I am suggesting the former because it can be done at one place (.targets) and works for all solutions. I am not against using libnosys, I just simply need GCC builds fixed.  (I can provide a PR with necessary changes in the scatterfiles to export the 'end' symbol and leave nosys.specs; it saves about 30 bytes).

For more detailed reference how to properly implement system calls for libnosys please have a look for example at http://www.atmel.com/Images/Frequently-Asked-Questions-4.9.3.26.txt.

smaillet-ms:
OK, for 30 bytes I can live with that fix. Something to note for vNext so we don't do that again (We'll have plenty more new mistakes to make :grin: )

josesimoes:
To point out that the these "end", "sbrk" and all the other missing symbols are all gone with #486.

PR #486, opened 2016-08-05
Remove custom heap

josesimoes:
- remove all references and code related with custom heap
- rework scatter files, AppEntry, FirstEntry and TinyHAL
- reworked start up files to perform memory initializations
- clean-up unused code
- implements #473

I've tested this on real hardware with the Discovery4 solution.
The binaries sizes are: Tinybooter.bin 41372B & ER_FLASH  304880B
I've successfully build the MCBSTM32F400 but can't test it on hardware.
As I'm not entirely confortable nor can test it, I have NOT changed the scatter files and start up assembly code for MDK, only for GCC. Appreciate if someone knowledgeable on that platform could step in and help with that.

Ready for the code review and the criticism. 👷 😓 😄


smaillet-ms:
Please remove the inclusion of additional PRs that makes it impossible to roll back or process PRs independently. Each PR should be isolated and focused on doing one thing that is as small and narrowly defined as possible.

Application/MicroBooter/MicroBooter.cpp
```diff
@@ -3,7 +3,7 @@
 //  <No description>
 //
 //  Microsoft dotNetMF Project
-//  Copyright �2004 Microsoft Corporation
+//  Copyright �2004 Microsoft Corporation
```
smaillet-ms: please do not alter copyright comments/notices in the code.

josesimoes:
Which PR are you referring to?

CLR/Core/CLR_RT_Memory.cpp
```diff
@@ -16,8 +16,6 @@ static int s_PreHeapInitIndex = 0;
 
 ////////////////////////////////////////////////////////////
 
-HAL_DECLARE_CUSTOM_HEAP( CLR_RT_Memory::Allocate, CLR_RT_Memory::Release, CLR_RT_Memory::ReAllocate );
```
smaillet-ms, line 19: Umm, hold on a sec... This is defining the managed heap APIs. The HAL_DECLARE_CUSTOM_HEAP creates a set of private_xxx APis that are inlined wrappers around the methods listed in the macro. While the entire approach to using a macro like this is dubious at best, we don't want to eliminate the functionality  as intended here. The use here is intended to be the managed heap which has lots of internal requirements and assumptions for the collector to work properly.

Application/MicroBooter/MicroBooter.cpp
josesimoes: To be accurate there is no change in the copyright notice. I haven't changed that obviously.
Being the copyright symbol that is different I would say that GitHub is having problems with the code page of the file it something like that.
But I'll do my best removing that difference from the file.

CLR/Graphics/Graphics.cpp
```diff
@@ -195,7 +195,7 @@ HRESULT CLR_GFX_Bitmap::CreateInstance( CLR_RT_HeapBlock& ref, const CLR_UINT8*
         /* When loading a Windows BMP, GIF, or JPEG file, it is converted in-place to the native BPP.
          * When loading a compressed TinyCLR Bitmap from a resource file, two bitmaps are needed to decompress, then convert.
          * This fragments the heap and wastes space until the next garbage collection is done.
-         * When using the SimpleHeap, there is no relocation, so decompressing into a temp bitmap into the simpleheap wastes
+         * When using the heap, there is no relocation, so decompressing into a temp bitmap into the heap wastes
```
smaillet-ms, line 198: The comment shouldn't just blindly replace words with a GREP, it needs to be re-worded to reflect the consequences. The platform creator should still have the option of creating a simple heap/big object storage to support large bitmaps that are otherwise not supported in the managed heap. (One of the implementation details in the managed heap is an assumption on the maximum size of any single allocation)

smaillet-ms:
@josesimoes I'm referning to this statement: 

> NOTE: included workaround for BIG_ENDIAN issue #481 
> Please don't mix PRs -  It's fine to state that one PR has a dependency on another one but don't include the changes for one in another.

josesimoes:
I can remove the statement. As for PR #481 it was just merged into the repo, so I guess it wouldn't even allow a merge if I removed the respective changes...

CLR/Core/CLR_RT_Memory.cpp
josesimoes, line 19: Are you entirely sure about that?!
The name refers to custom heap.
I have build this without this code and it went fine.
I have loaded the image on real hardware a run a test app. Creating and destroying objects. Checked the managed memory going up and down a seeing the GC kicking on. I admit it wasn't nothing fancy or exhaustive but it proved that the managed heap was working.

CLR/Graphics/Graphics.cpp
josesimoes, line 198: I'm not following you here... A platform creator can make the standard heap as large as he wants to hold these graph and whatever stuff it's needed. I don't see why this comment is misleading or not providing enough explanations for an informed decision.

smaillet-ms:
Now that it is merged you can re-base so hopefully those changes will disappear and we can focus on the heap specific changes.

josesimoes:
@smaillet-ms done!

CLR/Graphics/Graphics.cpp
smaillet-ms, line 198: The large object heap was isolated for a reason, to allow the OEM to partition out memory used for LARGE objects. Throwing everything into one big heap may be a simple solution to port but can lead to fragmentation problems, which is why the large buffer and bitmap images were isolated into a separate heap. This might be something an OEM could manage using a custom implementation of malloc()/free() (e.g. use the size parameter of malloc to determine which underlying region to allocate from), but doing so is more difficult as replacing the standard functions is slightly less than intuitively obvious to the casual observer 😉 and compiler/tool specific. While I'm all for getting rid of custom heap use for the normal PAL/HAL work (that should stick with standard malloc for greater portability and re-use) I think it makes life easier for ports needing large buffers to still have this as an option.

josesimoes, line 198: My suggestion was to remove the custom heap. Which I did, for GCC. And I suspect it should be easy for someone knowledgeable to port this to MDK. 

Wasn't aware of this nuance of "large objects". Feel free to take what's done and incorporate any changes required to making the options you mention available.

Application/MicroBooter/MicroBooter.cpp
```diff
@@ -3,7 +3,7 @@
 //  <No description>
 //
 //  Microsoft dotNetMF Project
-//  Copyright �2004 Microsoft Corporation
+//  Copyright ©2004 Microsoft Corporation
```
josesimoes: @smaillet-ms I've replaced this with the correct copyright symbol. Saved the file with UTF-8 encoding and GitHub still shows a difference at this line. 
If you really think this deserves the effort let me the encoding for the original code page and I'll try to revert this particular change.

miloush: @josesimoes The original encoding of the file is Windows 1252. (ISO 8859-1 might work too)

josesimoes: @miloush just saved it again with that encoding and seems to have remove that diff. Thanks for pointing that. 👏

smaillet-ms: Ugh! :face_palm: Sigh... Thanks for finding this. Unfortunately fixing the encoding on mass is likely to cause more problems than it will help. Given that the current plan on [vNext](https://github.com/NETMF/netmf-interpreter/tree/v5_dev) is to start with a clean slate so we can also do code style clean ups as we essentially import code into a new branch we can define and normalize the encoding for each file type. (And ideally find a way to detect deviations before the get into the repo)

miloush: Yeah, I was thinking with the vNext in mind whether there is still any tool in the chain that couldn't handle Unicode...

smaillet-ms: Embedded Assemblers? C? C++? Who knows what sort of shortcuts some of the available compilers use. (I can't remember if C++ officially supports Unicode text or even if it makes any mention of encoding of source files in any particular form, EBCDIC anyone? 😁 )

piwi1263: @smaillet-ms EBCDIC I eat that for breakfast, it's my daily job ....

josesimoes:
This solves the need for #492

PR #494, opened 2016-08-11
Update Install-BinTools.ps1

ghost:
"The easiest way to install the required binary tools is to run the Install-BinTools.ps1 PowerShell Script from the root of your repository"
the old version installed the tools as sub from the root of the repo what is wrong as it has to be sibling of the root of the repo


smaillet-ms:
I need to check on how this impacts the automated builds as this was created for that purpose and documented later for local use. Hopefully All I have to do is adjust the directory it runs in (and possibly the order of things as it may try to install the bin tools before it creates the repo folder, I'll need to review that) Either way any impacts should be manageable on the automated build server side.

smaillet-ms:
This script is not broken. It is important to note that it uses the $SPOROOT variable (set by the imported module .\tools\scripts\Build-netmf.psm1) $SPOROOT is the top level directory for all of the NETMF build which contains the tools, bin and the source code folders. $SPOCLIENT is the source code folder most people deal with everyday.  Since the tools and bin files are located under $SPOROOT the script in it's original form was correct. @cyberh0me  What problems were you seeing where it wasn't working correctly?

PR #512, opened 2016-09-07
Add gitignores for Visual Studio Code

josesimoes:
From GitHub official repository (see repo github/gitignore/Global/VisualStudioCode.gitignore)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,12 @@
 ipch/
 *.VC.db
 
+# Visual Studio Code
+.vscode/*
+!.vscode/settings.json
+!.vscode/tasks.json
+!.vscode/launch.json
+
 !/tools/bin/
 
 #uVision
--- a/.gitignore
+++ b/.gitignore
@@ -33,9 +33,6 @@ ipch/
 
 # Visual Studio Code
 .vscode/*
-!.vscode/settings.json
-!.vscode/tasks.json
-!.vscode/launch.json
 
 !/tools/bin/
 


smaillet-ms:
👍

cw2:
Just out of curiosity, why the .json files needs to be present in the repo?

I am not against the change, I have already excluded .vscode/ in my branch, but I certainly have different settings.json and I don't use Code debugger (so no launch.json). These are user-specific files...

josesimoes:
@cw2 great question... as I've mentioned: I took that from the official repo. Didn't gave it much thought, I admit. I guess that's the way to save the user's preferences and settings back to the repo. Actually I don't think this is of use to NETMF in particular so those 3 lines can be there or not.

smaillet-ms:
User specific settings should be ignored in the repo. However VS code (and .NET CORE) have used .json files as the project file format - though from what I have read they are moving away from that as it has not been a positive experience. `<Soapbox>`It effectively would became a new MSBUILD with a different syntax and entirely new tooling that effectively replicated what was already available in MSBuild`</Soapbox>`  I can't remember the details on which file was used for which purpose and there's not much in the docs at this point. At this point I'd prefer it if none of the VSCode .json files are included in the repo as VSCode isn't officially supported, is still undergoing changes and vNext will be changing the build system. So I can only see trouble and confusion.

.gitignore
```diff
@@ -31,6 +31,12 @@
 ipch/
 *.VC.db
 
+# Visual Studio Code
+.vscode/*
+!.vscode/settings.json
```
smaillet-ms: I mis-read this earlier, let's not be including these in the tree as the VSCode system is still in flux and not officially supported (e.g. just block all of the .vscode folder and leave it at that. Otherwise people are forced to use VSCode to ensure the JSON files are updated correctly when making changes even if they don't otherwise use VSCode. By allowing them into the tree we effectively make them officially part of NETMF and they must then be maintained tested and validated for each PR etc...

Issue #521, opened 2016-10-25
build_solution for GCC


josesimoes:
I'm failing to use the current version of the build_solution to build a solution with GCC.

It seems to require an extra argument in 3rd line "call setenv_%1 %2 **_%3**_" like this:

```
@echo off
setlocal
call setenv_%1 %2 %3
cd Solutions\%4
msbuild /flp:verbosity=detailed /clp:verbosity=minimal %~5
endlocal&&exit /B %ERRORLEVEL%

```

can anyone please double check this and also confirm if this causes any issues with MDK?

Thanks!

smaillet-ms:
No idea - what branch are you based off of? The [current dev branch version of build_solution  ](https://github.com/NETMF/netmf-interpreter/blob/dev/build_solution.cmd)doesn't have that. 

The automated build calls build_solution with the following args for the MDK build:
`MDK 5.05 MCBSTM32F400`

For GCC it uses an additional parameter to specify the project to build (relative to the solution folder):
`GCC 4.9.3 MCBSTM32F400 TinyClr_nonet\Tinyclr_nonet.proj`
and again with:
`GCC 4.9.3 MCBSTM32F400 TinyClr_nonet\Tinyclr_nonet.proj`
to validate the known GCC builds complete (which they don't , which is why the build status indicates partial success)
The additional and optional 4th parameter allows targeting a specific MSBuild project, if the extra parameter is not present then MSBuild will use the project file in the solution directory. If more then one exists it returns an error indicating that there is more than one and you need to specify which one you want.

josesimoes:
@smaillet-ms yes, the current version does not have that and it fails to build a solution based with GCC. That's why I've raised this question. I'm on the dev branch.

I can only make it work with that 3rd parameter, otherwise it fails complaining with an invalid argument

Compiler: GCC 5.4.1
setting vars for GCC compiler GCC5.4.1

Error - Invalid argument (C:\Program Files (x86)\GNU Tools ARM Embedded\4.9 2015q1).  Usage: setenv_GCC.cmd GCC_VERSION GCC_TOOL_PATH
        Example:  setenv_gcc.cmd 4.2.1 c:\tools\gcc

A correct call has to be something like:
build_solution gcc 5.4.1 E:\GNU_Tools_ARM_Embedded\5_4_2016q2 STM32F4DISCOVERY

smaillet-ms:
I think the better fix is in setenv_base:
 in this area:

```
rem use a default for GCC

[...]

PR #523, opened 2016-10-26
Update script to dequote parameter

josesimoes:
(see #521)

### DIFF
--- a/build_solution.cmd
+++ b/build_solution.cmd
@@ -1,6 +1,6 @@
 @echo off
 setlocal
-call setenv_%1 %2
+call setenv_%1 %~2
 cd Solutions\%3
 msbuild /flp:verbosity=detailed /clp:verbosity=minimal %~4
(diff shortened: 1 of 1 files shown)


build_solution.cmd
```diff
@@ -1,6 +1,6 @@
 @echo off
 setlocal
-call setenv_%1 %2
+call setenv_%1 %~2
 cd Solutions\%3
 msbuild /flp:verbosity=detailed /clp:verbosity=minimal %~4
-endlocal&&exit /B %ERRORLEVEL%
```
smaillet-ms, line 6: Why did this line change? Seems unnecessary.

josesimoes: That line hasn't changed. I've edited the file directly on GitHub. What that icon at the end is telling you is that the file is missing the empty line that Github requires to make file comparisons.

smaillet-ms: Interesting, never seen that one before...

PR #526, opened 2016-11-07
Add suport to use hardware CRC calculation for F4 series

josesimoes:
- add native code project for F4 series
- update TinyBooter and TinyCLR projects in Solutions to be able to choose from hardware/software CRC library

This provides a nice performance improvement when debugging and making use of the debug connection which CRC calculations abundantly on every packet rx/tx.

### DIFF
--- a/DeviceCode/Targets/Native/STM32F4/DeviceCode/STM32F4_CRC/CRC_functions.cpp
+++ b/DeviceCode/Targets/Native/STM32F4/DeviceCode/STM32F4_CRC/CRC_functions.cpp
@@ -30,65 +30,117 @@ typedef struct
     crc: previous CRC value to start CRC computing
 */
 
-/* NOTES: 
-    - should be good for F1, L1, F2 and F4 units series, other have diferent configurations and polynomial coefficients
-    - can only handle buffer sizes multiple of 4 otherwise the calculation is incorrect
-    - can not handle initial CRC values, only 0 is admited
-    - all the above can be fixed with more code so feel free to jump in ;)     
-*/
+/* NOTE: is good for F1, L1, F2 and F4 units series, other have diferent configurations and polynomial coefficients */
 
+static UINT32 ReverseCRC32(UINT32 targetCRC)
+{
+    // nibble lookup table for _REVERSE_ CRC32 polynomial 0x4C11DB7
+    static const UINT32 crc32NibbleTable[16] =
+    {
+        0x00000000, 0xB2B4BCB6, 0x61A864DB, 0xD31CD86D, 0xC350C9B6, 0x71E47500, 0xA2F8AD6D, 0x104C11DB,
+        0x82608EDB, 0x30D4326D, 0xE3C8EA00, 0x517C56B6, 0x4130476D, 0xF384FBDB, 0x209823B6, 0x922C9F00
+    };
+    UINT8 counter = 8;
+
+    while(counter--)
+    {
+        targetCRC = (targetCRC >> 4) ^ crc32NibbleTable[targetCRC & 0x0F];
+    }
+
+    return targetCRC;
+}
+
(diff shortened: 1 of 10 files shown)


DeviceCode/Targets/Native/STM32F4/DeviceCode/STM32F4_CRC/CRC_functions.cpp
```diff
@@ -0,0 +1,94 @@
...
+    size: lenght of buffer to compute CRC
+    crc: previous CRC value to start CRC computing
+*/
+
+/* NOTES: 
+    - should be good for F1, L1, F2 and F4 units series, other have diferent configurations and polynomial coefficients
+    - can only handle buffer sizes multiple of 4 otherwise the calculation is incorrect
+    - can not handle initial CRC values, only 0 is admited
+    - all the above can be fixed with more code so feel free to jump in ;)     
+*/
```
smaillet-ms: Given these restrictions how does this work with NETMF, NETMF often chunks data into blocks to maintain the appearance of multitasking in a single thread. So it will compute the CRC for a portion of data, use that result as the starting seed of the next so if only 0 is allowed, won't that fail?

It would be interesting to understand the variations of hardware CRC so that we can adjust the HAL layer API surface to better accommodate real world hardware, if not in the interpreter at least it could be done for AOT compilation.

DeviceCode/Targets/Native/STM32F4/DeviceCode/STM32F4_CRC/CRC_functions.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+
+UINT32 SUPPORT_ComputeCRC(const void* buffer, int size, UINT32 initCrc)
```
smaillet-ms, line 68: Do you have any performance measurements on the hardware based solution vs. software? The table based software approach is pretty small in code size, big on data size for the table and reasonably fast, so I'm curios to see how much better the hardware form is.

DeviceCode/Targets/Native/STM32F4/DeviceCode/STM32F4_CRC/CRC_functions.cpp
josesimoes: The current working Solutions are for F4 series which are covered by this code.
I thought on adding this note there to make people aware of the hardware differences with other STM32 series. 
The hardware variations are documented in ST's [AN4187](http://www.st.com/en/embedded-software/stsw-stm32an4187.html)

DeviceCode/Targets/Native/STM32F4/DeviceCode/STM32F4_CRC/CRC_functions.cpp
josesimoes: I haven't hard measure it. But being the CRC peripheral in the silicon it would be expected to be way faster than any software algorithm. 
Anyway you can read all about ST own comparison on their App note above. They claim the hardware peripheral is about 60 times faster than the software equivalent.

DeviceCode/Targets/Native/STM32F4/DeviceCode/STM32F4_CRC/CRC_functions.cpp
cw2: @smaillet-ms Unfortunately, the hardware CRC unit design on F4 is rather limited, further iterations support configurable polynomial, 8/16/32 bit data size, programmable initial value and bit reversal, which makes it significantly easier to use.

The calculation takes 4 AHB clock cycles for 32-bit data; the data size can be dynamically adjusted to minimize the number of accesses (i.e. 32, 16, 8 for 7 bytes).

josesimoes:
I believe that this PR now has all the features of the original implementation...

Issue #212, opened 2015-06-25
VB debug static variable examination causes crash


rockybooth:
I have been debugging a large VB program and having lots of debug issues and decided to try to find the cause. With the following not so long program, I can crash Visual Studio (2012 or 2013) every time by simply waiting for the stop, then using the cursor to see the present value of I. Using the immediate window to find the value will also crash VB. 

Local variables declared inside the sub can be explored without issue but variables passed to a sub or function also cannot be examined in that sub or function. 

Similar code in C# using Static instead of Private works just peachy.

This happens with the emulator, so this is not a GHI issue, and none of their dlls are loaded.

Option Explicit On
Option Strict On

Imports Microsoft.SPOT
Imports Microsoft.VisualBasic.CompilerServices

Namespace MFConsoleApplication1

    Public Module Module1
        Private I As Integer = 1

        Sub Main()

            Debug.Print(Resources.GetString(Resources.StringResources.String1))
            Stop
            Debug.Print(I.ToString)
        End Sub

    End Module

End Namespace

 
Presumably the issue is in Microsoft.VisualBasic.dll
This occurs running 
Microsoft.Spot.Native 4.3.1
mscorlib 4.3.1
Microsoft.Visualbasic 1.0.0.0 (v4.0.30319)

I hope I am posting in the right place.  Also see https://www.ghielectronics.com/community/forum/topic?id=19600&page=1#msg193443
where this is discussed

smaillet-ms:
Thanks for reporting this, we'll investigate and post what we find here. Yes, this is the right place to post bugs and other discussions on the .NET Micro Framework.

rockybooth:
Can anyone comment on when this might be resolved or if there are workarounds?
Thanks

smaillet-ms:
Other than not using statics we don't have any workarounds yet. We're working on core runtime and porting issues first (such as a memory leak in OpenSSL)

We're more than happy to have someone look into this, if someone is interested feel free to claim this one here.

martincalsyn:
I have a fix for this which I will submit as a PR once I clean it up a bit.  The changes are all in the CorDebug component. Basically CorDebugClass has to return an instance of CorDebugGenericType in the CorDebugClass implementation of ICorDebugClass2.GetParameterizedType.  And CorDebugGenericType needs an implementation for GetStaticFieldValue.  Those two changes stop the crashing and will return proper hover and watch values for VB globals.

rockybooth:
[...]

PR #551, opened 2018-09-20
Issue212 fix

martincalsyn:
This fixes Issue #212 where VS crashes when hovering over a global var in a vb module.

The  fix here is to make sure that  CorDebugClass responds correctly to calls to CorDebugClass.GetParameterizedType. Originally it should have returned E_NOTIMPL, but it returned OK and a null which is what causes the crash.  This function is supposed to return full specified types derived from generic classes and wasn't implemented because netmf doesn't do generics, but it also gets called for classes with no specialization types when trying to create a type enclosure for Vb globals.

Now, with this change, it returns a properly specialized type definition.  The constructor had to be specialized to accommodate the fact debugger values known the appdomain, and type specializations need an assembly and type domain.


martincalsyn:
Ugh - not at all what I intended to do

PR #552, opened 2018-09-20
Fix #212 - Hovering over Vb globals crashes VS

### LINKED ISSUE #212: VB debug static variable examination causes crash

martincalsyn:
This fixes Issue #212 where VS crashes when hovering over a global var in a vb module.

The fix here is to make sure that CorDebugClass responds correctly to calls to CorDebugClass.GetParameterizedType. Originally it should have returned E_NOTIMPL, but it returned OK and a null which is what causes the crash. This function is supposed to return full specified types derived from generic classes and wasn't implemented because netmf doesn't do generics, but it also gets called for classes with no specialization types when trying to create a type enclosure for Vb globals.

Now, with this change, it returns a properly specialized type definition. The constructor had to be specialized to accommodate the fact debugger values known the appdomain, and type specializations need an assembly and type domain.


Framework/CorDebug/CorDebugClass.cs
```diff
@@ -122,7 +122,7 @@ public uint TypeDef_Index
         int ICorDebugClass2.GetParameterizedType( CorElementType elementType, uint nTypeArgs, ICorDebugType []ppTypeArgs, out ICorDebugType ppType )
         {
             // CorDebugClass.GetParameterizedType is not implemented
```
martincalsyn, line 124: Technically, I guess this comment isn't true anymore.  It is implemented, but only works for non-generic types (which is the only kind of type that ever gets requested)

